PR #3, opened 2016-07-12
Initial version of publish/download specs

noahfalk:
There are known changes based on feedback that are still to come,
but this should get us a baseline.


noahfalk:
@tmat - ptal

PR #4, opened 2016-07-12
Adding a sample symbol server implementation

noahfalk:


noahfalk:
@tmat - ptal

src/samples/NugetSymbolServer/NugetSymbolServer.xproj
```diff
@@ -0,0 +1,21 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="14.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <PropertyGroup>
+    <VisualStudioVersion Condition="'$(VisualStudioVersion)' == ''">14.0</VisualStudioVersion>
+    <VSToolsPath Condition="'$(VSToolsPath)' == ''">$(MSBuildExtensionsPath32)\Microsoft\VisualStudio\v$(VisualStudioVersion)</VSToolsPath>
```
tmat: Can we avoid dependency on VS?

noahfalk: You can build it with dotnet which has no VS dependency. I've also removed the project from the solution so that it won't build by default.

We'll need to figure out how to get a .csproj for it, but following the existing template didn't work for this ASP.NET app. I haven't spent much time trying to figure out the scope of the problem and how to work around it.

noahfalk:
Updated to not be nested in the samples folder and corrected the doc link.
@tmat - ptal

tmat:
:+1:

PR #5, opened 2016-07-12
Add the parsers, indexer, and downloader

noahfalk:
This has initial implementations of a lot of the functionality we need, but the projects still need some massaging to be a good citizens in the repo.


noahfalk:
@tmat - ptal

src/FileFormats.ELF/ELFFile.cs
```diff
@@ -0,0 +1,288 @@
...
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+
+namespace FileFormats.ELF
+{
+
+    public class ELFFile
+    {
+        IAddressSpace _dataSource;
```
tmat: ```private readonly``` on all these fields?

src/FileFormats.ELF/ELFFile.cs
```diff
@@ -0,0 +1,288 @@
...
+            List<ELFSegment> segments = new List<ELFSegment>();
+            for (uint i = 0; i < Header.ProgramHeaderCount; i++)
+            {
+                segments.Add(new ELFSegment(DataSourceReader, _position, 
+                    _position + Header.ProgramHeaderOffset + i*Header.ProgramHeaderEntrySize, _isDataSourceVirtualAddressSpace));
+            }
+            return segments;
+        }
+
+        Reader CreateVirtualAddressReader()
```
tmat: ```private```

src/FileFormats.ELF/ELFFile.cs
```diff
@@ -0,0 +1,288 @@
...
+                        }
+                    }
+                }
+            }
+
+            return null;
+        }
+
+        #region Validation Rules
+        public ValidationRule IsHeaderPhOffValid
```
tmat: What is "PhOff"?

tmat:
We need to convert .xproj to .csproj files. I'll take care of that, for now let's just avoid adding these xproj files into the .sln so that they don't get built.

noahfalk:
@tmat - ptal
I needed to update the xproj to csproj in order to run the codeformatter tool. I had to edit the projects by hand because VS didn't want to add references, but it seems to read and build them OK. I think this addresses all the feedback + all the todo items that would be useful before checkin.

tmat:
Editing by hand is currently necessary, until we get a better project system in VS.

src/FileFormats/FileFormats.csproj
```diff
@@ -0,0 +1,33 @@
...
+    <AssemblyName>FileFormats</AssemblyName>
+    <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
+    <TargetFrameworkVersion>v5.0</TargetFrameworkVersion>
+    <TargetFrameworkProfile>
+    </TargetFrameworkProfile>
+    <TargetFrameworkIdentifier>.NETPortable</TargetFrameworkIdentifier>
+    <NuGetTargetMoniker>.NETStandard,Version=v1.5</NuGetTargetMoniker>
+  </PropertyGroup>
+  <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Debug|AnyCPU' ">
+    <NoWarn>1570;1591</NoWarn>
```
tmat: Probably want these in Release build as well.

tmat:
:+1:

PR #9, opened 2016-07-15
Update the samples for the spec changes

noahfalk:
Also reverted the samples back to portable console apps so that they run properly


noahfalk:
@tmat -ptal

PR #10, opened 2016-07-19
Restoring the xproj build

noahfalk:
The build system in this repo has been problematic to work with ; ) I made SymStore_dotnet.sln as an alternate option that is more useful for local development right now.

I have tried getting the test projects to run following the pattern of the existing unit tests but I have not been successful. FileFormats needs a netstandard 1.5 dependency to build and then the test projects need the same dependency. Once I do that I can no longer follow the existing template exactly, though its unclear if this is really the source of the problem. Ultimately the tests do not show up in the VS test window which is awkward for development. However to get the tests working properly in the xproj based projects I needed to reference netcoreapp instead of netstandard 1.5, and that breaks the build from the .csproj build system. I removed the test projects from SymStore.sln so that CI doesn't break.

For the console apps I tried using a non-portable console app template, but trying to reference to ..\..\build\Targets\Imports.targets to match the other projects immediately broke the build. It seemed that if projects were going to use different sets of rules I may as well stay consistent and use the .xproj/dotnet scheme rather than introducing a 3rd variation.

I can easily imagine there is a better, more consistent solution, but it will take more time or someone more knowledgeable in this area to find it.

### DIFF
--- a/SymStore.sln
+++ b/SymStore.sln
@@ -13,23 +13,13 @@ Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Microsoft.SymbolStore.Clien
 EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "FileFormats", "src\FileFormats\FileFormats.csproj", "{256F3AFA-B659-44B0-BC6F-3C9912E9DD4F}"
 EndProject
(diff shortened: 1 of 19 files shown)


noahfalk:
@tmat

tmat:
Using netstandard 1.5 shouldn't be a problem. I use TestDriven.NET to run the tests and it works ok. Moving it out of the main solution is fine for now, if it helps you to develop locally.

PR #11, opened 2016-07-19
Update unit test project

noahfalk:
VS insistently wants to change this project every time I open the solution.
I had been repeatedly undoing it, but its easier to let VS make its desired edit
than to keep undoing it.


noahfalk:
@tmat

PR #13, opened 2016-07-21
EmbedIndex improvements

noahfalk:
1) It now accepts directory arguments to handle all the packages inside as a batch
2) It warns of ELF files that are missing build ids
3) It now properly indexes .so files


noahfalk:
@tmat

PR #14, opened 2016-08-04
bug fixes to ELFCoreFileReader to fix VA offset issues, and rebasing …

schaabs:
…issues


schaabs:
@noahfalk, these are the updates we made while debugging through this code.  Also I changed ReadLoaded images to only return images which live at the lowest base offset.

PR #16, opened 2016-08-04
Build an EmbedIndex tool package during the build

dagood:
Restores/publishes using the dotnet tool, independent from the "main" build. Ideally it would use existing artifacts from the main build, but it creates its own. The NuGet package is generated normally, with a new nuspec.

@noahfalk @tmat

### DIFF
--- a/build/Targets/Imports.targets
+++ b/build/Targets/Imports.targets
@@ -7,5 +7,4 @@
   <Import Project="..\Toolset\DefaultImports.targets"/>
   <Import Project="..\Toolset\GenerateAssemblyInfo.targets" />
   <Import Project="..\Toolset\GenerateInternalsVisibleTo.targets" />
-  <Import Project="..\Toolset\DotNetPublish.targets" />
 </Project>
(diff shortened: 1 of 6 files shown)


dagood:
Ah, well those test failures seem to explain why the main solution doesn't include EmbedIndex--looks like CLI isn't installed. I can download it during the build, unless that sounds unreasonable.

tmat:
Yes, feel free to download it.

build/Targets/Imports.targets
```diff
@@ -7,4 +7,5 @@
   <Import Project="..\Toolset\DefaultImports.targets"/>
   <Import Project="..\Toolset\GenerateAssemblyInfo.targets" />
   <Import Project="..\Toolset\GenerateInternalsVisibleTo.targets" />
+  <Import Project="..\Toolset\DotNetPublish.targets" />
```
tmat: I'd prefer if this Import was directly in the EmbedIndex project rather then global for all projects.

dagood:
Thanks, made those changes and got CI to pass.

noahfalk:
LGTM

PR #18, opened 2016-08-18
Implement FileFormat.Minidump, Microsoft.SymbolServer.Client

leculver:
- Added Minidump support to FileFormats (with basic functionality).
- Added support for PEFile being in the virtual address space.
- Added an initial implementation of a symbol sever client.  Note that portable pdbs, linux binaries/symbols, and mac binaries/symbols are not yet implemented.
- Added more tests to existing functionality and the new functionality.  Note that I have added about 100 megs worth of test data...while this isn't great, I think it's more important in the short term to have working test data until we can get an azure link up that it's downloadable from.


leculver:
Currently build.cmd is not functional.  I'm working to resolve that before checkin, but I'd still like code review feedback on this.

src/FileFormats.Minidump.Tests/Tests.cs
```diff
@@ -0,0 +1,142 @@
...
+
+            // We should only read the header and not too far into the dump file, 1k should be plenty.
+            Assert.True(readHelper.Max <= 1024);
+        }
+
+        private Stream GetCrashDump(string path)
+        {
+            MemoryStream ms = new MemoryStream();
+            using (FileStream fs = File.OpenRead(path))
+            using (GZipStream gs = new GZipStream(fs, CompressionMode.Decompress))
```
noahfalk: nit: indent

src/FileFormats.Minidump.Tests/Tests.cs
```diff
@@ -0,0 +1,142 @@
...
+                gs.CopyTo(ms);
+            return ms;
+        }
+        
+        private static Minidump GetMinidumpFromStream(Stream stream)
+        {
+            IAddressSpace sas = new StreamAddressSpace(stream);
+            Minidump minidump = new Minidump(sas);
```
noahfalk: Feel free to add overloads to the constructors for the file format parsers that take a Stream directly if you want. I imagine this two line construction would be a common pattern and happy to take that down to just one line.

src/FileFormats.Minidump/Minidump.cs
```diff
@@ -0,0 +1,155 @@
...
+        {
+            return _dataSourceReader.WithAddressSpace(new MinidumpVirtualAddressSpace(Segments, _dataSource));
+        }
+
+        private List<MinidumpLoadedImage> CreateLoadedImageList()
+        {
+            if (_moduleListStream == -1)
+                throw new BadInputFormatException("Minidump does not contain a ModuleStreamList in its directory.");
+            
+            MINIDUMP_MODULE[] modules = _dataSourceReader.ReadCountedArray<MINIDUMP_MODULE>(_directory[_moduleListStream].Rva);
```
noahfalk: This wouldn't handle a non-zero position I think? Rather than remembering to add _position everywhere you can also create a new RelativeAddressSpace represents a portion of the original, and use that as the address space of the _dataSourceReader

src/FileFormats.Minidump/MinidumpSegment.cs
```diff
@@ -0,0 +1,54 @@
...
+        /// Returns whether the given address is contained in this region of virtual memory.
+        /// </summary>
+        /// <param name="address">A virtual address in the orginal process's address space.</param>
+        /// <returns>True if this segment contains the address, false otherwise.</returns>
+        public bool Contains(ulong address)
+        {
+            return VirtualAddress <= address && address < VirtualAddress + Size;
+        }
+
+        internal static MinidumpSegment Create(MINIDUMP_MEMORY_DESCRIPTOR region)
```
noahfalk: Any reason not to use a constructor instead?

src/FileFormats.Minidump/MinidumpStructures.cs
```diff
@@ -0,0 +1,163 @@
...
+                     .AddPointerTypes()
+                     .AddNullTerminatedString()
+                     .AddTStructTypes();
+        }
+    }
+
+#pragma warning disable 0649
+#pragma warning disable 0169
+
+    internal class MINIDUMP_HEADER : TStruct
```
noahfalk: I try to stick with BCL casing conventions that would name this MinidumpHeader (though to be fair I am not sure if I even consistently did that in my own code)

src/FileFormats.Minidump/MinidumpStructures.cs
```diff
@@ -0,0 +1,163 @@
...
+    {
+        public const int MINIDUMP_VERSION = 0x504d444d;
```
noahfalk: pascal casing again

src/FileFormats.Minidump/MinidumpStructures.cs
```diff
@@ -0,0 +1,163 @@
...
+        PROCESSOR_ARCHITECTURE_IA64 = 6,
+        PROCESSOR_ARCHITECTURE_ALPHA64 = 7,
+        PROCESSOR_ARCHITECTURE_MSIL = 8,
+        PROCESSOR_ARCHITECTURE_AMD64 = 9,
+        PROCESSOR_ARCHITECTURE_IA32_ON_WIN64 = 10,
+    }
+
+    internal sealed class VS_FIXEDFILEINFO : TStruct
+    {
+        public uint dwSignature;            /* e.g. 0xfeef04bd */
```
noahfalk: I'd get rid of the hungarian

src/FileFormats.Minidump/MinidumpStructures.cs
```diff
@@ -0,0 +1,163 @@
...
+    }
+
+
+    internal sealed class MINIDUMP_LOCATION_DESCRIPTOR : TStruct
+    {
+        public uint DataSize;
+        public uint Rva;
+    }
+
+    [TStructPack(4)]
```
noahfalk: The naming for other attributes like ArraySize didn't prefix with TStruct, so I would suggest simply [Pack(4)]

src/FileFormats.Minidump/MinidumpVirtualAddressSpace.cs
```diff
@@ -0,0 +1,75 @@
...
+        public uint Read(ulong position, byte[] buffer, uint bufferOffset, uint count)
+        {
+            if (count == 0)
+                return 0;
+            
+            MinidumpSegment seg = FindSegment(position);
+            if (seg == null)
+                return 0;
+            
+            // TODO: What if they read past the end of the segment?
```
noahfalk: There is a PiecewiseAddressSpace that I hope already handles this correctly.

src/FileFormats.PE/PEFile.cs
```diff
@@ -28,13 +29,14 @@ public class PEFile
         private const int PESignatureOffsetLocation = 0x3C;
         private const uint ExpectedPESignature = 0x00004550;    // PE00
 
-        public PEFile(IAddressSpace fileAddressSpace)
+        public PEFile(IAddressSpace fileAddressSpace, ulong position = 0)
         {
+            _position = position;
             _fileAddressSpace = fileAddressSpace;
```
noahfalk: Rather than convert all the addresses to '_position + X' you can define the _fileAddressSpace = new RelativeAddressSpace(fileAddressSpace, position)

src/FileFormats/Reader.cs
```diff
@@ -22,6 +23,30 @@ public Reader(IAddressSpace dataSource, LayoutManager layoutManager)
         public LayoutManager LayoutManager { get; private set; }
         public IAddressSpace DataSource { get; private set; }
 
+        public string ReadCountedString(ulong position, Encoding encoding)
```
noahfalk: I'm hoping to leave Reader very simple, otherwise I worry it will fill with too many ReadXYZ() helpers that are scenario specific (for example what if another scenario has a counted string with a 2 byte length instead of 4?) As an alternative I would suggest either:
a) Create a static class with extension methods, callers wouldn't need to change at all
b) Create a type CountedStringLayout that is similar to NullTerminatedStringLayout. 
     i) Only one layout can be used when you call reader.Read<string>( ... ), but you can choose which one it is when you set up your LayoutManager
     ii) OR you could create a type CountedString that has an implicit cast to string and do string s = reader.Read<CountedString>( ... )

Right now there isn't much advantage to doing the layout unless you particularly like the Read<T>(...) syntax, however in the future you could imagine something like a TStruct or array that composes over types that aren't fixed size. I'd probably go with the static extension method for now, but either one works.

src/FileFormats/Reader.cs
```diff
@@ -22,6 +23,30 @@ public Reader(IAddressSpace dataSource, LayoutManager layoutManager)
...
+        {
+            uint elementCount = Read<uint>(ref position);
+            byte[] buffer = Read(position, elementCount);
+            return encoding.GetString(buffer);
+        }
+
+        public T[] ReadCountedArray<T>(ulong position)
```
noahfalk: Same for CountedArray as CountedString

src/Microsoft.SymbolStore.Client.Tests/SymbolLocatorTests.cs
```diff
@@ -0,0 +1,74 @@
...
+namespace Microsoft.SymbolStore.Client
+{
+    public partial class Tests
+    {
+        [Fact]
+        public async void SymbolServerEndToEnd()
+        {
+            SymbolLocator locator = new SymbolLocator(new SymbolCache(CacheLocation), new ISymbolServer[] { CreateWindowsSymbolServer() });
+
+            Task<string>[] tasks = new Task<string>[2];
```
noahfalk: How about something like this?

Task peTask = (async () => ValidatePEFile(locator, await locator.FindPEFileAsync(PEFilename, PEFileTimestamp, PEFileSize).Result))();
Task pdbTask = (async () => { ValidatePdb(locator, await locator.FindPdbAsync(PDBFileName, PDBGuid, PDBAge).Result))();
await Task.WhenAll(peTask, pdbTask);

src/Microsoft.SymbolStore.Client/CabStream.cs
```diff
@@ -0,0 +1,265 @@
...
+        private IntPtr Notify(NotificationType type, Notification notification)
+        {
+            if (type != NotificationType.CopyFile)
+                return IntPtr.Zero;
+
+            switch (type)
+            {
+                case NotificationType.CopyFile:
+                    _outputHandle = CreateStreamPointer(_output);
+                    Debug.WriteLine($"Notify({type}, {notification}) - {ToHex(_outputHandle)}");
```
noahfalk: Can we take these Debug.WriteLines out or put them behind a conditional? I worry that people debugging an app that happens to use this class will need to deal with this spew but it won't relate to the thing they are trying to debug.

src/Microsoft.SymbolStore.Client/FileSemaphore.cs
```diff
@@ -0,0 +1,87 @@
...
+                    {
+                    }
+                    
+                    _disposed = true;
+                }
+            }
+            
+            ~LockedFile()
+            {
+               Dispose(false);
```
noahfalk: Rather than disposing the file I'd suggest asserting. I can't imagine its ever correct to forget to call Dispose on this type explicitly. (Doing stuff in finalizers is also a pain, for example I think its legal for _lockPath to already be GC'ed at this point)

src/Microsoft.SymbolStore.Client/FileSemaphore.cs
```diff
@@ -0,0 +1,87 @@
...
+        }
+
+        public static IDisposable TryLockFile(string fullPath)
+        {
+            DirectoryInfo di = new FileInfo(fullPath).Directory;
+            if (!di.Exists)
+                throw new DirectoryNotFoundException(di.FullName);
+
+            string lockName = fullPath + ".sem";
+            while (true)
```
noahfalk: This while loop appears unneeded

src/Microsoft.SymbolStore.Client/StoreQueryBuilder.cs
```diff
@@ -17,14 +19,29 @@ internal static class StoreQueryBuilder
...
         {
-            return fileName + "/" + PortablePdbPrefix + guid.ToString("N") + stamp.ToString("x8") + "/" + fileName;
+            Debug.Assert(filename == Path.GetFileName(filename));  // parts of the path should have already been stripped
+            Debug.Assert(age >= 0);
+
+            string result = $"{filename}/{PortablePdbPrefix + guid.ToString("N") + age.ToString()}/{filename}";
+            return result;
+        }
+
+        public static string GetWindowsPdbQueryString(string filename, Guid guid, int age)
```
noahfalk: Its probably more accurate to call this a 'Microsoft', 'Microsoft C++' or 'MSF' pdb.

src/Microsoft.SymbolStore.Client/StoreQueryBuilder.cs
```diff
@@ -17,14 +19,29 @@ internal static class StoreQueryBuilder
 
         public static readonly string SourcePrefix = "sha1";
```
noahfalk: The sha1 indexing scheme can be used for any purpose, it just happens that currently its being used for source. 

The prefix is also "sha1-"

src/Microsoft.SymbolStore.Client/SymbolCache.cs
```diff
@@ -0,0 +1,160 @@
...
+            if (string.IsNullOrWhiteSpace(location))
+                throw new ArgumentException($"Argument '{nameof(location)}' cannot be an empty string.");
+
+            _location = location;
+            if (!Directory.Exists(_location))
+                Directory.CreateDirectory(location);
+        }
+
+        #region PDB Functions
+        public string GetPdbFromCache(string pdbName, Guid guid, int age)
```
noahfalk: If the cache isn't frequently used directly by consumers of the library, it may not be worthwhile creating a strongly typed set of Get/Store/Remove/GetCacheLocation methods for every file type. Instead you can could create a generic set of methods that takes a relative path and let the caller calculate it for you.

src/Microsoft.SymbolStore.Client/SymbolLocator.cs
```diff
@@ -0,0 +1,162 @@
...
+using System.Linq;
+using System.Net;
+using System.Threading.Tasks;
+using FileFormats;
+
+namespace Microsoft.SymbolStore.Client
+{
+    public sealed class SymbolLocator
+    {
+        private static string[] s_microsoftSymbolServerUrls = new string[] { "http://msdl.microsoft.com/download/symbols", "https://nuget.smbsrc.net", "http://referencesource.microsoft.com/symbols", "https://dotnet.myget.org/F/dotnet-core/symbols" };
```
noahfalk: This line should be broken up (it scrolls right off the side of my fairly high resolution monitor)

src/Microsoft.SymbolStore.Client/SymbolLocator.cs
```diff
@@ -0,0 +1,162 @@
...
+        private static ISymbolServer[] s_microsoftSymbolServers = s_microsoftSymbolServerUrls.Select(url => new WindowsSymbolSever(url)).Cast<ISymbolServer>().ToArray();
+
```
noahfalk: (Its late, needed to stop here for the night - I'll try to finish reviewing as soon as I can)

src/FileFormats.Minidump.Tests/Tests.cs
leculver: This is now the preferred way of doing indentation of multiple using statements.  Visual Studio enforces this and I'd rather not fight VS when coding...so that has to stay.

src/FileFormats.Minidump.Tests/Tests.cs
leculver: There's potentially a lot of places where we'd have a second overload.  I actually think FileFormats should expose a Stream extension method "AsAddressSpace" which does that.  I'm not going to do it in this checkin, so we don't have to debate here.

src/FileFormats.Minidump/Minidump.cs
leculver: I actually wonder if we shouldn't remove the position argument entirely from both Minidump and ElfFile/etc.  Instead the caller should wrap everything with RelativeAddressSpace...that seems like a cleaner and safer way than doing it all manually.  Not a change I'm going to make in this PR, so I'll fix the offset for now.

src/FileFormats.Minidump/MinidumpStructures.cs
leculver: One issue you are going to run into here is you've defined TStruct and friends as public.  "Pack" is such a common term that I worry you'll run into collisions with "PackAttribute", hence why I named it something off-key.  However, I suppose that's why namespaces exist...I've changed it.

src/FileFormats.Minidump/MinidumpVirtualAddressSpace.cs
leculver: I suppose, but I worry about the number of IAddressSpace instances I would need to create...  Similarly, you didn't use piecewise address space in Elf either...ELFVirtualAddressSpace.

I also worry your implementation (both in piecewise and elf) does a linear search for every read...that's bound to get slow, but something that's fixable.

src/Microsoft.SymbolStore.Client.Tests/SymbolLocatorTests.cs
leculver: Yeah I dunno what I was thinking there.  Instead of using async lambdas I just changed ValidatePEFile/ValidatePdb to async and doing all the work.

src/Microsoft.SymbolStore.Client/CabStream.cs
leculver: These were not meant to stay in the final code, I'll strip them out.

src/Microsoft.SymbolStore.Client/FileSemaphore.cs
leculver: I'll do both...  You can't count on the user doing the right thing in all cases, so we need the finalizer to clean up, but we should also assert it never happens.

src/Microsoft.SymbolStore.Client/StoreQueryBuilder.cs
leculver: This was in the original code Tomas checked in.  I'm going to leave it alone for now, and revisit when I actually go to use it.

src/Microsoft.SymbolStore.Client/StoreQueryBuilder.cs
leculver: I needed something to call it as opposed to "PortablePDB" which will also be here.  Windows PDB vs Portable PDB, seems to be the right split...rather than Microsoft PDB vs Portable PDB (when they are both microsoft formats...).

We'll talk in person on that one at some point.

src/Microsoft.SymbolStore.Client/SymbolCache.cs
leculver: This allows me to finely control the location and layout of all files.  This actually has a very useful property when we end up using the result.  Allowing the user to control where files could go (by accepting a path string, and someone modifying what we return as the index path) would make the rest of the code more difficult..

src/FileFormats.Minidump.Tests/Tests.cs
noahfalk: Interesting. I agree, if this (IMO odd) indenting is the convention VS encourages then lets leave it as is.

src/FileFormats.Minidump.Tests/Tests.cs
noahfalk: To be clear, you'll not get a debate from me if you don't do it all - merely a suggestion : )

src/FileFormats.Minidump/Minidump.cs
noahfalk: The position argument actually matters when you are reading a loaded image from the memory space. You need to determine whether or not you loaded at the preferred base address or an alternate address in order to handle any  encoding or API that uses VAs.

src/FileFormats.Minidump/MinidumpVirtualAddressSpace.cs
noahfalk: The current ELF implementation was written before I had created the PiecewiseAddressSpace. I agree the linear search isn't ideal, but it could easily be replaced with a binary search as you have done here. I'm not opposed to checking this in as-is, but it seems like in the long run we don't need 3 or 4 different implementations of VirtualAddressing that only differ in terms of what file format data structure held the mapping info.

src/Microsoft.SymbolStore.Client/SymbolLocator.cs
```diff
@@ -0,0 +1,169 @@
...
+using FileFormats;
+
+namespace Microsoft.SymbolStore.Client
+{
+    public sealed class SymbolLocator
+    {
+        private static readonly string[] s_microsoftSymbolServerUrls = new string[]
+        {
+            "http://msdl.microsoft.com/download/symbols",
+            "https://nuget.smbsrc.net",
```
noahfalk: >"https://nuget.smbsrc.net" [](start = 12, length = 26)

This is just my curiosity, what is this one? I had never heard of it before.

src/Microsoft.SymbolStore.Client/SymbolLocator.cs
```diff
@@ -0,0 +1,169 @@
...
+        }
+        
+        public async Task<string> FindPdbAsync(string pdbName, Guid guid, int age)
+        {
+            string fullPath = pdbName;
+            pdbName = Path.GetFileName(pdbName);
+
+            // Does the PDB file exist on disk with the right criteria?  If so use it and don't archive
+            // it in the cache.  (This scenario is for developers hitting F5 on their machine.)
+            if (fullPath != pdbName && File.Exists(fullPath))
```
noahfalk: >fullPath != pdbName && File.Exists(fullPath)) [](start = 16, length = 45)

I'd rather not add this here. In my mental model of the layering I imagine two different levels of functionality:
a) The ability to get files from symbol servers (possibly cached)
b) The ability to get files from anywhere they can be found using a variety of strategies such as local search paths, registry and environment variable configuration, breadcrumbs embedded in other files, being adjacent to a known file, well known paths, symbol server, etc. This is usually filled with tons of configuration and policies depending on the exact scenario/tool that is using it.

I viewed SymbolLocator as being (a), and I view this particular code as being in (b). I absolutely think (b) is a useful thing and we could make class that helps people do it, I just don't think it should go here. Many tools probably already have their own implementation of (b) and we shouldn't deny them the ability to search exactly the places they want to if symbol server is the only thing they want.

src/Microsoft.SymbolStore.Client/SymbolLocator.cs
```diff
@@ -0,0 +1,169 @@
...
+                return Cache.StorePEFile(result.Stream, filename, timestamp, filesize);
+            
+            CabConverter converter = new CabConverter(result.Stream);
+            using (MemoryStream stream = await converter.ConvertAsync())
+                return Cache.StorePEFile(stream, filename, timestamp, filesize);
+        }
+
+        private async Task<SymbolServerResult> SearchServers(Func<ISymbolServer, Task<SymbolServerResult>> getTask)
+        {
+            IEnumerable<Task<SymbolServerResult>> tasks = _symbolServers.Where(server => server.PreferThisServer).Select(server => getTask(server));
```
noahfalk: >server => getTask(server) [](start = 121, length = 25)

Nit: you should be able to simplify 'server => getTask(server)' to 'getTask'

src/Microsoft.SymbolStore.Client/SymbolLocator.cs
```diff
@@ -0,0 +1,169 @@
...
+        
+        async Task<T> GetFirstNonNullResult<T>(List<Task<T>> tasks) where T : class
+        {
+            while (tasks.Count > 0)
+            {
+                Task<T> task = await Task.WhenAny(tasks);
+
+                T result = task.Result;
+                if (result != null)
+                    return result;
```
noahfalk: >return result; [](start = 20, length = 14)

Not needed for checkin, but adding cancelation support might make some nice perf improvements in the future. As I understand the current algorithm we initiate a download from multiple servers. Even after one of them is finished we continue to download from all the other servers in the background and then allow their work to be GC'ed without using it.

src/Microsoft.SymbolStore.Client/WindowsSymbolSever.cs
```diff
@@ -0,0 +1,234 @@
...
+        {
+            try
+            {
+                string path = _path;
+                Uri uri = new Uri(path);
+                string host = uri.DnsSafeHost;
+
+                if (!_isServer)
+                    throw new InvalidOperationException("Cannot use IsReachable if this is not a remote server.");
+                Task<IPAddress[]> addresses = System.Net.Dns.GetHostAddressesAsync(uri.DnsSafeHost);
```
noahfalk: This seemed like a surprising implementation of IsReachable. I assumed it would test if you could get network traffic from the remote host but this appears to only check if the hostname is resolvable via DNS. What do you think of trying to send the host a message and make sure we get something back?

src/Microsoft.SymbolStore.Client/WindowsSymbolSever.cs
```diff
@@ -0,0 +1,234 @@
...
+            _path = path;
+            _isServer = isServer;
+        }
+
+        public bool IsReachable()
+        {
+            return IsReachable(700);
+        }
+
+        public bool IsReachable(int timeout)
```
noahfalk: >timeout [](start = 36, length = 7)

I would use TimeSpan or change the parameter name to imply what the unit of time is.

src/Microsoft.SymbolStore.Client/WindowsSymbolSever.cs
```diff
@@ -0,0 +1,234 @@
...
+            {
+                string path = _path;
+                Uri uri = new Uri(path);
+                string host = uri.DnsSafeHost;
+
+                if (!_isServer)
+                    throw new InvalidOperationException("Cannot use IsReachable if this is not a remote server.");
+                Task<IPAddress[]> addresses = System.Net.Dns.GetHostAddressesAsync(uri.DnsSafeHost);
+
+                if (!addresses.Wait(700))
```
noahfalk: I think this was supposed to be 'timeout'

src/Microsoft.SymbolStore.Client/WindowsSymbolSever.cs
```diff
@@ -0,0 +1,234 @@
...
+        bool IsReachable();
+        bool IsReachable(int timeout);
+        
+        SymbolServerResult FindPEFile(string filename, int buildTimeStamp, int imageSize);
+        Task<SymbolServerResult> FindPEFileAsync(string filename, int buildTimeStamp, int imageSize);
+        SymbolServerResult FindPdb(string pdbName, Guid guid, int age);
+        Task<SymbolServerResult> FindPdbAsync(string pdbName, Guid guid, int age);
+    }
+
+    public class WindowsSymbolSever : ISymbolServer
```
noahfalk: I'd like to propose doing some refactoring, but probably easier to chat in person. I have some different ideas to bounce off you and we can come to some agreement about what happens pre-checkin vs. later. Thanks for putting this all together Lee!

src/Microsoft.SymbolStore.Client/SymbolLocator.cs
leculver: Honestly I just pulled it from perfview, I'm not sure myself.  :)

src/Microsoft.SymbolStore.Client/SymbolLocator.cs
leculver: Fair point.  I've changed that.

src/Microsoft.SymbolStore.Client/SymbolLocator.cs
leculver: That's not quite how it works...  ISymbolServer only produces a Stream + metadata about that stream.  It does **not** read from the stream.  Our SearchServers/GetFirstNonNullResult will simply wait for the first symbol sever to say "I have this file, here is the stream".  If any other symbol sever produces a result, then we will discard that second stream and it will be GC'ed without ever having been read from.

Adding cancellation is tricky, since HttpWebRequest doesn't accept a cancellation token, and while the HttpWebRequest.Abort method can be used, the implementation is a little tricky.  Since we never read from the stream, we _should_ be ok to not cancel here...but if that turns out to not be the case we can certainly revisit this.

src/Microsoft.SymbolStore.Client/WindowsSymbolSever.cs
leculver: This was meant as a more 'does this server exist' method.  In the sense that tools we (CLR) write will want to check whether symweb is available or not, but actually it might not even be appropriate to put that in the library.  It might be a better choice to simply leave it to the consumer of this library and strip out IsReachable altogether.

src/Microsoft.SymbolStore.Client/SymbolLocator.cs
noahfalk: I think I tracked it down. It appears to the endpoint for downloading nuget packages hosted by symbolsource.org. I doubt it should be in this list.

src/Microsoft.SymbolStore.Client/WindowsSymbolSever.cs
noahfalk: I was doing some experiments with alternate implementations/refactorings of the SymbolServer here:
https://github.com/noahfalk/symstore/commit/0bbd743582e45b9798196d28bf85c25ea159b6a8

Its incomplete and some parts of it I'd certainly change, but here are the parts I think could be interesting:
a) SymbolCaches are a sub-type SymbolStore. All SymbolStores let you find files, SymbolCaches also let you add files.
b) I added the concept of a FileIdentity. You can give a key to any SymbolStore and it will give you back an identity. Symbol caches are allowed to assume that if two searches produce the same FileIdentity, then they will have the same content. In an ideal world the key itself would be a good identity, but in reality different servers give different content for the same key (ie symweb). I used the URI of the file on a particular server as the FileIdentity but that isn't required.
c) I created two implementations of the cache. One works like the current caches which ignores FileIdentity and allows you to store at most one item with the same key. The other can store content per-FileIdentity, even when there is more than one FileIdentity used for the same key. This lets you re-use the same cache for different symbol servers without any loss of fidelity about what came from where. Currently its not back-compatible with the old cache layout, but it could be made compatible by arbitrarily choosing one FileIdentity as the primary one for each key and storing it in the expected location. It also makes too many copies of files, but something like hard-links could probably remedy the needless duplicates.
d) Once you have a cache that can hold more than one file per-key, you can do file selection after caching rather than before. The UnionSymbolServer does it on a first response basis, but you could also rank sources in priority order or examine the content that came back to decide what is the best one to use.
e) I dealt almost entirely in terms of opaque keys. I added a few extension methods that offer convenience routines to consume the strongly typed file identifiers, but I would guess 95+% of the logic is file format agnostic. I assumed the keys were in a/b/c format, but even that doesn't seem like a hard requirement. I know you were predicting working with arbitrary keys was going to make it harder, but I'm not sure I follow your concerns on that one. From my experimenting so far I came to the opposite conclusion.

PR #29, opened 2017-04-30
Upgrade to VS2017 and new dotnet .csproj files

noahfalk:
Get rid of our previous SxS independent builds for dotnet/VS2015 and replace with a unified build for both.


noahfalk:
This is going to need the CI portion of the build scripts to be upgraded to use dotnet.exe. I'll come back to this at some point.

japj:
I believe this is already in master with commit 7264cb93857da6d00692b965d45f00fef32d0106?

PR #32, opened 2017-05-17
Add indexing of symbol catalog files.

crummel:
This shows two possible approaches for adding symbol catalog indexing to EmbedIndex:
1. Add either *.cat or a particular catalog filename or path to the SHA-1 indexer.
2. Add a symbol catalog indexer that indexes the catalog separately.

The SHA-1 indexing approach is simple and the code already exists, but I'm not sure it satisfies security requirements.  The symbol catalog indexer uses SHA-256 which should be fine security-wise, and additionally marks a file as a catalog instead of a generic file, which could have some advantages.

I considered a third approach, similar to the symbol files where a more natural ID is used.  The catalog file itself has internal IDs or signatures that we could use as IDs, but I think this opens up new issues:
- The built-in catalog ID is only 16 bytes and as far as I know, is not generated securely.
- Using a hash of the file list has the advantage of using a hash already appearing in the catalog, but as it does not include the signatures, is vulnerable to cache poisoning with an incorrectly-signed file.
- Several signatures (of file hash list, entire catalog, timestamp info) appear in the catalog and should be unique, but are 512 bytes long, which seems excessive when we can simply use a hash of the whole file instead.
- All of these would require bringing DerSequenceReader in to EmbedIndex and parsing the catalog, which could take a bit more development time.

Overall I'm favoring the separate catalog indexer approach.

### DIFF
--- a/src/EmbedIndex/Program.cs
+++ b/src/EmbedIndex/Program.cs
@@ -64,6 +64,7 @@ namespace EmbedIndex
                 using (ZipArchive archive = new ZipArchive(packageStream, ZipArchiveMode.Update))
                 {
                     List<Tuple<string, string>> indexEntries = ComputeIndexEntries(archive, indexers);
+                    indexEntries.AddRange(ComputeCatalogEntries(archive, indexers));
 
                     ZipArchiveEntry symbolIndexEntry = archive.GetEntry("symbol_index.json");
                     if (symbolIndexEntry != null)
@@ -161,6 +162,35 @@ namespace EmbedIndex
                 }
             }
             return keys;
+        }
+
+        private static IEnumerable<Tuple<string, string>> ComputeCatalogEntries(ZipArchive archive, IEnumerable<IFileFormatIndexer> indexers)
+        {
+            foreach (ZipArchiveEntry catalogEntry in archive.Entries)
+            {
+                if (!catalogEntry.Name.ToLowerInvariant().EndsWith(".cat"))
+                {
+                    continue;
+                }
+                string signedFile = catalogEntry.FullName.Substring(0, catalogEntry.FullName.Length - 4);
+                ZipArchiveEntry signedFileEntry = archive.Entries.SingleOrDefault(e => e.FullName == signedFile);
+                if (signedFileEntry == null)
+                {
+                    continue;
+                }
+                foreach (IFileFormatIndexer indexer in indexers)
+                {
+                    using (Stream signedFileStream = signedFileEntry.Open())
+                    {
+                        string key = indexer.ComputeIndexKey(signedFile, signedFileStream);
+                        if (key != null)
+                        {
+                            key = key.Substring(0, key.LastIndexOf("/") + 1) + "signature.cat";
+                            yield return Tuple.Create(catalogEntry.FullName, key);
+                        }
+                    }
+                }
+            }
         }
     }
 }


crummel:
@noahfalk @dagood Please take a look, if you agree on the approach I'll remove the SHA-1 changes and update the docs.

noahfalk:
I think the overall scenario will still have a gap with this approach so I'm going to propose a modification. For the issue, consider that you are a developer that has a dump file. You've got binaries and symbols downloaded already but you want to verify their authenticity, so to do that you need to download the catalog file. At this point the only things you know are information about the files to be verified, but you don't have any information about the catalog file. In particular for this scheme you do not know the hash of the catalog file. Thus you are unable to format an appropriate URL to download it.

In terms of solving that gap I could see two options:
1) Index the catalog solely using information from the other files. For example if there is a pdb we want to validate that has guid 123456789 and age 1, we might make a key foo.pdb/1234567891/signatures.cat. This means we have to index the catalog file N different times with N unique keys, but all of them point back to the same catalog file.
2) Similar to option 1 but use an indirection. In this case we create a 2nd file in the package whose entire contents is just the hash for the catalog. For example we call the file catalog_index_id.txt and its contents is the string "1458697828423123451233", assuming that is a hash of the catalog in question. Now we again add N unique indexing entries of the form foo.pdb/1234567891/catalog_index_id.txt + the entry you proposed for signatures.cat/sha1-1458697828423123451233/signatures.cat. At the client end you need two requests to grab the file, 1 to look up the hash and then a 2nd to use that hash to download the catalog.

I'm thinking the indirection may be the better way to go because it allows the client to avoid re-downloading the same signature.cat N seperate times. The client has no way in advance to know that foo.pdb/cat-12345/files.cat and bar.pdb/cat-54321/files.cat are actually the same file, so with the first plan it has to do N full downloads. The latency is arguably higher with plan 2 because it requires two network round trips in the worst case, but in practice I assume these downloads will be running concurrently with acquiring the binaries/pdbs which will dominate the overall response time to acquire everything.

In terms of hashes I don't think you need to worry about their security properties. The presumption is that a symbol server is insecure and that the content you get back could be poisoned. The catalog is self-validating via the signature inside it. The worst case is that someone does collide the indexing hash, you get back a malicious catalog file, and then the client side tooling detects it is bad and refuses to use it. This results in a denial of service, but not an escalation of privilege. That said, it should be pretty easy to use SHA-256 and since there is no pre-existing constraint forcing use of a weaker hash (as there was for source files) we may as well use a strong one. Mostly that is to avoid any unintended connotation of weak security, despite the hash not serving a security need.

dagood:
I agree that of those two, the indirection sounds better. Since we most likely have to download something for each file anyway, have individual catalogs already been ruled out? Is there significant per-catalog overhead?

 3. Create one signed catalog per file in the symbol package. The indexing entry is the same as the one for the file itself, but named `signature.cat`. E.g. for `foo.pdb/1234567891/foo.pdb`, it would be called `foo.pdb/1234567891/signature.cat`.

crummel:
From my notes, I have that catalog scaling is pretty consistently about 190 bytes per file, plus 360 bytes overhead, plus 8KB for the signature.  So that would take us to more like 8.6KB per file, which doesn't seem like it would be a concern size-wise (the symbol package we're using for testing is 15MB and has five symbol files).  The other concern would be how long signing takes but as long as we batch them instead of signing one-by-one we at least should only take as long as the product repo does for signing.  @noahfalk?

crummel:
I pushed a commit to use Davis' approach.  I'm not in love with how it turned out - I'm going to try doing the catalog indexing as a separate step instead of inside the other indexers next.  I am wondering how much we want to verify the catalog upfront - I'm trying to avoid bring the actual catalog parsing in, but without it we would probably just be writing out the catalog index entries for *.dbg instead of verifying that a specific .dbg file actually exists in the catalog.  Is this a concern?

crummel:
I like this one a bit better (real set of changes [here](https://github.com/dotnet/symstore/pull/32/commits/4dda1eb609a3130d13531f6bd7bbd31c3ec8c3e4#diff-81657814ddeba5c0d185e35e2441bae8)), but it is very dependent on the key format.  I'll look at what it would take for a single catalog and for some up-front verification now.

noahfalk:
My instinct was to merge them, but I believe you are right - I should resist that urge and keep it simple. One catalog per-file is simpler than sharing them and outside the case of numerous trivially small binaries/pdbs the overhead really isn't that much. If we later discover the performance isn't good enough we can always invest in a more performant option at that point.

dagood:
I'm curious about the extra effort it might take to add the "follow indirection" feature into the client if we defer it. That 8.6 KB per file does add up--not very significant in terms of symbol package size, but debugger responsiveness seems like it could suffer a lot earlier.

The Core-Setup symbol archive zip for linux has 142 index-worthy files in it. That brings individual catalogs to ~1200 KB, vs 35 KB for merged. (That's not including DLLs, native binaries, and sources.) But that's only if someone needs to download all 142 files. Seems hard to say if it's worth it or not.

noahfalk:
How big are those 142 files? Assuming its something like 25-50MB, +2-4% overhead seems modest for the debugger and I wouldn't worry about it in the v1 implementation.

dagood:
They're 205 MB. :smile: For some reason I've been considering a situation where someone needs to verify all the files but didn't download them in the same action--I can't come up with a scenario where that happens. Sounds good.

crummel:
I've updated with just the single-file-per-catalog approach and doing the catalog indexing as a separate step.  My remaining concern is that this is dependent on the format of the indexing keys, do you think this is a problem @noahfalk?

noahfalk:
I'm fine assuming that keys will have the forward slash. LGTM, thanks Chris

PR #33, opened 2017-05-22
Fix symbol catalog key name.

crummel:
- Key name/blob path tuple was switched around
- Lowercase key name
- Use catalog file name instead of "signature.cat"


crummel:
@noahfalk This should fix it.

crummel:
cc @dagood

PR #34, opened 2017-06-01
Update to use one-catalog-per-nupkg mode.

crummel:
- also simplify/unify how we calculate hashes.

### DIFF
--- a/src/EmbedIndex/PDBSignatureAgeIndexer.cs
+++ b/src/EmbedIndex/PDBSignatureAgeIndexer.cs
@@ -32,7 +32,7 @@ namespace EmbedIndex
                 key.Append(filename);
                 key.Append("/");
                 key.Append(pdb.Signature.ToString().Replace("-", "").ToLowerInvariant());
-                key.Append(pdb.Age.ToString("x").ToLowerInvariant());
+                key.Append(pdb.Age.ToString("x"));
                 key.Append("/");
                 key.Append(filename);
                 return key.ToString();
--- a/src/EmbedIndex/PETimestampSizeIndexer.cs
+++ b/src/EmbedIndex/PETimestampSizeIndexer.cs
@@ -30,8 +30,8 @@ namespace EmbedIndex
                 StringBuilder key = new StringBuilder();
                 key.Append(filename);
                 key.Append("/");
-                key.Append(reader.Timestamp.ToString("x").ToLowerInvariant());
-                key.Append(reader.SizeOfImage.ToString("x").ToLowerInvariant());
+                key.Append(reader.Timestamp.ToString("x"));
+                key.Append(reader.SizeOfImage.ToString("x"));
                 key.Append("/");
                 key.Append(filename);
                 return key.ToString();
--- a/src/EmbedIndex/Program.cs
+++ b/src/EmbedIndex/Program.cs
@@ -5,6 +5,7 @@ using System.Collections.Generic;
 using System.IO;
 using System.IO.Compression;
 using System.Linq;
+using System.Security.Cryptography;
 
 namespace EmbedIndex
 {
@@ -12,18 +13,18 @@ namespace EmbedIndex
     {
         public static void Main(string[] args)
         {
-            if(args.Length != 2)
+            if (args.Length != 2)
             {
                 PrintUsage();
                 return;
             }
 
-            if(Directory.Exists(args[0]))
+            if (Directory.Exists(args[0]))
             {
                 string inputDir = args[0];
                 string outputDir = args[1];
                 Directory.CreateDirectory(outputDir);
-                foreach(string nugetPath in Directory.EnumerateFiles(inputDir, "*.nupkg", SearchOption.AllDirectories))
+                foreach (string nugetPath in Directory.EnumerateFiles(inputDir, "*.nupkg", SearchOption.AllDirectories))
                 {
                     Console.WriteLine(nugetPath);
                     string indexedNugetPath = Path.Combine(outputDir, Path.GetFileName(nugetPath));
@@ -127,7 +128,7 @@ namespace EmbedIndex
                 writer.WriteLine("    {");
                 writer.WriteLine("        \"clientKey\" : \"{0}\",", entry.Item1);
                 writer.WriteLine("        \"blobPath\" : \"{0}\"", entry.Item2);
-                
+
                 if (i != indexEntries.Count - 1)
                 {
                     writer.WriteLine("    },");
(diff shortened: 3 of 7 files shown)


crummel:
@noahfalk @dagood

src/EmbedIndex/Utility.cs
```diff
@@ -0,0 +1,19 @@
+﻿using System;
```
dagood: Missing copyright header.

src/EmbedIndex/Program.cs
```diff
@@ -164,33 +165,85 @@ public static IEnumerable<string> ComputeIndexEntries(string archiveRelativePath
...
                 }
-                string signedFile = catalogEntry.FullName.Substring(0, catalogEntry.FullName.Length - 4);
-                ZipArchiveEntry signedFileEntry = archive.Entries.SingleOrDefault(e => e.FullName == signedFile);
-                if (signedFileEntry == null)
+            }
+
+            // get list of cataloged symbol files
+            ZipArchiveEntry catalogedFilesEntry = archive.GetEntry(CatalogedSymbolListFileName);
+
+            if (catalogedFilesEntry == null)
```
dagood: Do we want to embed the `catalog_index_id.txt` even if `catalogedFilesEntry == null`? Might want to move this check to the beginning of the method.

src/EmbedIndex/Program.cs
```diff
@@ -164,33 +165,85 @@ public static IEnumerable<string> ComputeIndexEntries(string archiveRelativePath
...
+                        string key = indexer.ComputeIndexKey(fileEntry.FullName, signedFileStream);
                         if (key != null)
                         {
-                            key = key.Substring(0, key.LastIndexOf("/") + 1) + catalogEntry.Name.ToLowerInvariant();
-                            yield return Tuple.Create(key, catalogEntry.FullName);
+                            key = key.Substring(0, key.LastIndexOf("/") + 1) + CatalogIndexFileName.ToLowerInvariant();
+                            Tuple<string, string> ret = Tuple.Create(key, CatalogIndexFileName);
+                            if (!seen.Contains(ret))
+                            {
+                                seen.Add(ret);
```
dagood: Can be shortened to `if (seen.Add(ret)) { yield return ret; }`. When does this happen, though?

src/EmbedIndex/Program.cs
```diff
@@ -164,33 +165,85 @@ public static IEnumerable<string> ComputeIndexEntries(string archiveRelativePath
...
+                                yield return ret;
+                            }
                         }
                     }
                 }
             }
+
+            // add entry for the catalog file itself
+            yield return Tuple.Create($"{SymbolCatalogFileName}/catalog-{catalogHash}/{SymbolCatalogFileName}".ToLowerInvariant(), SymbolCatalogFileName);
```
dagood: \<nit> Line too long for my taste, hard to read/review. I'd suggest linebreaks before `$` and `SymbolCatalogFileName`.

src/EmbedIndex/PDBSignatureAgeIndexer.cs
```diff
@@ -32,7 +32,7 @@ public string ComputeIndexKey(string path, Stream fileStream)
                 key.Append(filename);
                 key.Append("/");
                 key.Append(pdb.Signature.ToString().Replace("-", "").ToLowerInvariant());
-                key.Append(pdb.Age.ToString("x").ToLowerInvariant());
+                key.Append(pdb.Age.ToString("x"));
```
dagood: Agreed, I don't see when `ToLowerInvariant` would do anything here.

src/EmbedIndex/Utility.cs
crummel: Added.

src/EmbedIndex/Program.cs
crummel: Changed as suggested.

src/EmbedIndex/Program.cs
crummel: Changed as suggested.  I saw this in the Microsoft.CSharp package - the netcore50 and netstandard1.3 versions of Microsoft.CSharp.dll are indentical so without this we had two identical entries (since our catalog entries don't include the path the way symbol entries do).

src/EmbedIndex/Program.cs
crummel: Changed as suggested.

src/EmbedIndex/PETimestampSizeIndexer.cs
```diff
@@ -30,8 +30,8 @@ public string ComputeIndexKey(string path, Stream fileStream)
                 StringBuilder key = new StringBuilder();
                 key.Append(filename);
                 key.Append("/");
-                key.Append(reader.Timestamp.ToString("x").ToLowerInvariant());
```
noahfalk: I'm assuming you already confirmed that ToString("x") prints lower case by default?

src/EmbedIndex/Program.cs
```diff
@@ -164,33 +165,86 @@ public static IEnumerable<string> ComputeIndexEntries(string archiveRelativePath
...
+            }
+            catalogedFilesEntry.Delete();
+
+            // for each regular symbol index entry covered by a symbol catalog,
+            // add an entry for the catalog file
+            HashSet<Tuple<string, string>> seen = new HashSet<Tuple<string, string>>();
+            foreach (string catalogedFile in catalogedFiles)
+            {
+                ZipArchiveEntry fileEntry = archive.GetEntry(catalogedFile);
                 foreach (IFileFormatIndexer indexer in indexers)
```
noahfalk: Have you double checked that we only wind up generating one catalog index per file? Its not huge deal if we generated other ones by iterating all indexers (such as the SHA-1 source index), but it would be a little wasteful of space.

src/EmbedIndex/PETimestampSizeIndexer.cs
crummel: Yes, this is correct.

src/EmbedIndex/Program.cs
crummel: Making sure I understand - the concern is that multiple indexers would return a key for the same file and the catalog indexing would include altered versions of them both?  Currently we won't write out exact duplicates, but if two indexers return different keys for a file we will modify and write out both keys for the catalog indexing.  If this isn't the desired behavior it should be an easy fix, but I'd want to make sure we're satisfied with the current order of the indexers since it essentially becomes a priority list.

noahfalk: My recollection (that might be wrong) is that we didn't necessarily pass every file to every indexer when we did the initial indexing. This meant the indexers might be seeing more files in this second pass than they did the first time around, thus producing more keys. As long as the results look correct in practice though I'm not super worried if there is some obscure theoretical corner case input that has extra keys.

crummel: Ah, ok, that makes sense.  Yes, I don't think this is a problem currently.  I think it would be better to fix an issue like this in the signing code if it does come up, so we can say cataloged.txt is accurate.

crummel:
@noahfalk I don't have permissions to merge this, could you do it please?

PR #36, opened 2017-06-13
Upgrade build to new dotnet/csproj files

mikem8361:
Upgrade the projects and build to use VS2017 and the new dotnet .csproj files based on the roslyn common RepoToolset files.

https://github.com/dotnet/roslyn-tools/blob/master/docs/RepoToolset.md


build/SignToolData.json
```diff
@@ -0,0 +1,10 @@
+{
+  "sign": [
+    {
+      "certificate": "MicrosoftSHA2",
+      "strongName": "MsSharedLib72",
+      "values": [
```
tmat: >"values": [ [](start = 6, length = 11)

I assume we'll add binaries to sign later...

build/Versions.props
```diff
@@ -0,0 +1,34 @@
...
+    <MicrosoftNetCompilersVersion>2.0.1</MicrosoftNetCompilersVersion>
+
+    <!-- Tests -->
+    <MicrosoftNETTestSdkVersion>15.0.0</MicrosoftNETTestSdkVersion>
+    <XUnitVersion>2.2.0</XUnitVersion>
+    <XUnitRunnerVisualStudioVersion>2.2.0</XUnitRunnerVisualStudioVersion>
+
+    <!-- desktop msbuild driver -->
+    <VSWhereVersion>1.0.47</VSWhereVersion>
+    <XUnitRunnerConsoleVersion>2.2.0</XUnitRunnerConsoleVersion>
```
tmat: This shouldn't be needed. Since we are not building VS extensions or WPF/WinFormas apps dotnet cli should be able to build everything in the repo.

src/EmbedIndex/EmbedIndex.csproj
```diff
@@ -0,0 +1,16 @@
...
+  <ItemGroup>
+    <ProjectReference Include="..\FileFormats.ELF\FileFormats.ELF.csproj" />
+    <ProjectReference Include="..\FileFormats.MachO\FileFormats.MachO.csproj" />
+    <ProjectReference Include="..\FileFormats.PDB\FileFormats.PDB.csproj" />
+    <ProjectReference Include="..\FileFormats.PE\FileFormats.PE.csproj" />
+    <ProjectReference Include="..\FileFormats\FileFormats.csproj" />
+  </ItemGroup>
+
+  <!-- Workaround for https://github.com/Microsoft/msbuild/issues/1721. -->
+  <Import Project="..\Directory.Build.targets" Condition="'$(TargetFramework)' == ''" />
```
tmat: Not necessary for projects that do not multi-target. Also, should be fixed in dotnet cli 2.0.

build/SignToolData.json
mikem8361: Yes, once we know all the binaries.

build/Versions.props
mikem8361: What about from VS 2017? I "guessed" that I might need them to run the tests from the VS test explorer.

tmat:
👍

PR #37, opened 2017-07-05
Create packages using .NET Core msbuild tooling

dagood:
Instead of the nuspec, use `IsPackable`. It's set to false by default by the Roslyn tooling, so I turned it on for each project we need for EmbedIndex and for Microsoft.SymbolStore.Client. I adjusted the target frameworks to netstandard versions.

This works with `Build.cmd -pack`.

Each project has its own package, rather than EmbedIndex being a standalone tooling package--this matches how we would like to consume it as a library in e.g. BuildTools and Core-Setup.

A little build definition work will be needed to start using this flow, assuming https://devdiv.visualstudio.com/DevDiv/_build/index?definitionId=3583&_a=completed is still the def.

@crummel @noahfalk


src/Directory.Build.props
```diff
@@ -5,6 +5,11 @@
   <!-- This file is imported by all projects at the beginning of the project files -->
   <Import Project="$(RepoToolsetDir)Settings.props" />
 
+  <PropertyGroup>
+    <IncludeSymbols>true</IncludeSymbols>
```
tmat: What do we do with symbol packages?

tmat:
👍

src/Directory.Build.props
tmat: I'd suggest enabling Embedded PDBs instead to avoid dealing with symbol server.

tmat:
Re the build definition - you can copy SymReaderConverter build definition. It has all the right settings.

src/Directory.Build.props
dagood: Currently nothing, but I figure uploading them to MyGet for indexing might help someone debug one of the projects down the line.

tmat: Enable Embedded PDBs instead. Much better experience and requires no extra work.

tmat: Also, I'm updating RepoToolset so that you get SourceLink for free as well. ETA next week.

dagood: Cool, I'll try that. Sounds like it doesn't matter in this case, but embedded PDBs would increase file size, right?

tmat: Yes.

mikem8361: Doesn't using embedded PDBs kind of defeat the whole point of creating the symbol package? We want the actual runtime assemblies to be as small as possible but still allow us, CSS and eventually outside devs to get the symbols (via the myget symbol server) the runtime/app/core dump they are debugging.

tmat: Why do you want them as small as possible for these tools?

mikem8361: Sorry, I completely misunderstood this change. For some reason I thought these were the changes to properly create the symbol packages for runtime. So you are right embedded PDBs are just fine for the tools.

dagood: PDBs are already embedded in release mode later in this file, so I just removed this line.

dagood:
Here's a build of this with the publishing steps disabled: https://devdiv.visualstudio.com/DevDiv/_build/index?buildId=845607

mikem8361:
Can we merge this PR now?

dagood:
Sounds good to me. (I don't have permission.)

PR #48, opened 2018-04-13
Fixed the dotnet-symbols cli extension packaging

mikem8361:
Build with 2.1 Preview 1.

Move SymClient to github repo.

Add nuget.config with <clear /> so the global nuget configuration doesn't affect this project's restore.

VS's nuget manager works better with nuget.config.

Make reading the (PE) debug directory PDB info more robust in ELF core dumps.


build/Versions.props
```diff
@@ -7,20 +7,12 @@
...
     <UsingToolXliff>false</UsingToolXliff>
   </PropertyGroup>
-
-  <PropertyGroup>
-    <RestoreSources>
-      $(RestoreSources);
-      https://dotnet.myget.org/F/dotnet-buildtools/api/v3/index.json;
-      https://dotnet.myget.org/F/roslyn-tools/api/v3/index.json;
-    </RestoreSources>
-  </PropertyGroup>
```
tmat: Why move it to nuget.config?

mikem8361: So it works better with VS. Otherwise VS displays the global feeds.

src/dotnet-symbol/dotnet-symbol.csproj
```diff
@@ -3,16 +3,22 @@
     <OutputType>Exe</OutputType>
     <TargetFramework>netcoreapp2.0</TargetFramework>
     <NoWarn>;1591;1701</NoWarn>
+    <IsPublishable>true</IsPublishable>
+    <IsPackable>true</IsPackable>
     <PackAsTool>true</PackAsTool>
-    <ToolCommandName>dotnet-symbols</ToolCommandName>
+    <!-- The package version needs to be hard coded as a stable version so "dotnet install -g dotnet-symbols" works -->
+    <PackageVersion>1.0.0</PackageVersion>
```
tmat: >PackageVersion>1.0.0</PackageVersion> [](start = 5, length = 37)

This is very odd. How are we gonna publish nightly builds of this package if all of them have the same version?

build/Versions.props
mikem8361: The <clear /> keeps the project from depending the global state too.

src/dotnet-symbol/dotnet-symbol.csproj
mikem8361: Yes this is, but this is a special kind of "global" tool dotnet cli package which only seems to work with no "dev" or "beta" versions.  I'm not publishing this yet until some signing issues the cli team have with these global tool packages.  Hopefully I can also figure out a better way to version.

build/Versions.props
tmat: This is not how other repos do it though [1]. If VS doesn't display the right feeds we need to fix VS. 
What global state? 

[1] We do not use nuget.config because it can't be overridden by orchestrated/source build. Although this repo is not part of orchestrated build we ought to be consistent.

tmat: > To point to feed with a reference built as part of the orchestrated build?

Yes.

I think this might be actually fixed by upgrading to the SDK version of the RepoToolset (1.0.0-beta2-62810-01). Try the latest master of symreader-converter.

mikem8361: Maybe the problem I had with symreader-converter was an older version. It builds fine now with nothing in the machine or user global nuget.config.

PR #56, opened 2018-04-16
Fix Linux build

tmat:


tmat:
@mikem8361 The Linux build is failing because .sh files are not marked executable. Needs:

```
git update-index --chmod=+x file.sh
```

mikem8361:
Are you saying I needed to add the execute bit before commiting the *.sh files?

tmat:
I merged your PR. Now it fails to build Microsoft.FileFormats.csproj which targets `net45` and `netstandard1.5`. Any reason why we need to target `net45`. We can do so, that's not a problem, just wondering if we need to.

mikem8361:
Yes, it needs to build for net45 because the DotNet-SymUploader build task needs to build for net45.

tmat:
OK, I added `<UsingToolNetFrameworkReferenceAssemblies>true</UsingToolNetFrameworkReferenceAssemblies>` to fetch the package containing reference assemblies for netfx

mikem8361:
I'm working on the Linux failures.

PR #58, opened 2018-04-16
Fix tests on Linux

mikem8361:
Add build and test scripts for Linux.

### DIFF
--- a/src/Microsoft.SymbolStore/KeyGenerators/KeyGenerator.cs
+++ b/src/Microsoft.SymbolStore/KeyGenerators/KeyGenerator.cs
@@ -88,7 +88,9 @@ namespace Microsoft.SymbolStore.KeyGenerators
         /// <returns>key</returns>
         protected static SymbolStoreKey BuildKey(string path, string id, bool clrSpecialFile = false)
         {
-            string file = Uri.EscapeDataString(Path.GetFileName(path).ToLowerInvariant());
+            // The back slashes are changed to forward slashes because Path.GetFileName doesn't work on 
+            // Linux /MacOS if there are backslashes. Both back and forward slashes work on Windows.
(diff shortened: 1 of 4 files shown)


src/Microsoft.SymbolStore/KeyGenerators/KeyGenerator.cs
```diff
@@ -88,7 +88,9 @@ public virtual bool IsDump()
         /// <returns>key</returns>
         protected static SymbolStoreKey BuildKey(string path, string id, bool clrSpecialFile = false)
         {
-            string file = Uri.EscapeDataString(Path.GetFileName(path).ToLowerInvariant());
+            // The back slashes are changed to forward slashes because Path.GetFileName doesn't work on 
+            // Linux /MacOS if there are backslashes. Both back and forward slashes work on Windows.
+            string file = Uri.EscapeDataString(Path.GetFileName(path.Replace('\\', '/')).ToLowerInvariant());
```
tmat: I believe replacing `\` with `/` for `GetFileName` is the right way to go.

This is what I use in SymReader-Converter to convert local paths to URIs:

```C#
 var escapedPath = string.Join("/", path.Substring(file.Path.Length).Split(new[] { '/', '\\' }).Select(Uri.EscapeDataString));
```

So that matches you implementation here, except ToLowerInvariant. Not sure why we would need to lower-case the file name here. I would rather not change the casing.

mikem8361: Lowercasing the filename is part of the indexing spec.  Not exactly necessary because most (but probably not all) symbol servers treat the index as case-insensitive.  The code has always lower cased it and I don't want to change it now.

tmat: OK

tmat:
LGTM

PR #77, opened 2018-10-09
Change some Warnings to Verbose tracing.

mikem8361:
Also fixed a case for the download tools to display an error if the symbol file can't be found.

### DIFF
--- a/src/Microsoft.SymbolStore/KeyGenerators/ELFFileKeyGenerator.cs
+++ b/src/Microsoft.SymbolStore/KeyGenerators/ELFFileKeyGenerator.cs
@@ -127,7 +127,7 @@ namespace Microsoft.SymbolStore.KeyGenerators
             }
             catch (Exception ex) when (ex is InvalidVirtualAddressException || ex is BadInputFormatException)
             {
-                Tracer.Warning("ELF .gnu_debuglink section in {0}: {1}", _path, ex.Message);
+                Tracer.Verbose("ELF .gnu_debuglink section in {0}: {1}", _path, ex.Message);
             }
             return null;
         }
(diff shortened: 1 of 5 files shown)


src/Microsoft.SymbolStore/KeyGenerators/ELFFileKeyGenerator.cs
```diff
@@ -127,7 +127,7 @@ private string GetSymbolFileName()
             }
             catch (Exception ex) when (ex is InvalidVirtualAddressException || ex is BadInputFormatException)
             {
-                Tracer.Warning("ELF .gnu_debuglink section in {0}: {1}", _path, ex.Message);
+                Tracer.Verbose("ELF .gnu_debuglink section in {0}: {1}", _path, ex.Message);
```
tmat: >Verbose [](start = 23, length = 7)

Does this mean the binary is bad?

src/Microsoft.SymbolStore/KeyGenerators/FileKeyGenerator.cs
```diff
@@ -40,7 +40,7 @@ public override IEnumerable<SymbolStoreKey> GetKeys(KeyTypeFlags flags)
                     return generator.GetKeys(flags);
                 }
             }
-            Tracer.Warning("Invalid file type: {0}", _file.FileName);
+            Tracer.Verbose("Invalid file type: {0}", _file.FileName);
```
tmat: >Invalid [](start = 28, length = 7)

I'd also change the message to "Unknown file type".

mikem8361:
No. It just means that the optional link/path to the debug file isn’t there.  All of .NET Core binaries release that have been “stripped” should have this section.  The debug info is stripped out of the .so and put into the .so.dbg and an “link” to the .so.dbg put into the .so.

mikem

tmat:
I see.

---
In reply to: [428427727](https://github.com/dotnet/symstore/pull/77#issuecomment-428427727) [](ancestors = 428427727)

Issue #83, opened 2019-01-07
NugetSymbolServer sample instructions out of date


dagood:
Steps in [the sample NuGet symbol server readme](https://github.com/dotnet/symstore/blob/7bbc70933673d5c4e4720a70ba7f999f1a76dde5/src/NugetSymbolServer/README.md) don't work anymore. The project is a tool project now, and doesn't run with `dotnet run`, and simply changing the project type raises complaints about files not being copied to the `artifacts` dir.

I was trying to use it to stand up a quick symbol server to see if some non-online symbols would be matched up and used correctly against a snapshot, but wasn't able to make it work.

It also didn't seem to load some recent `runtime.win-x64.Microsoft.NETCore.App.<version>.symbols.nupkg` files that I stuck in a `nuget_feed` dir in artifacts, but that is probably another issue not worth looking at until some overall flow works.

mikem8361:
This sample symbol server really isn't supported.  I really don't have the bandwidth to fix this.  Github also flagged a reference in this project as a security issue too (Microsoft.AspNetCore.Mvc).

Do you have sometime to fix up this project, update the reference and the README?

dagood:
> Do you have sometime to fix up this project, update the reference and the README?

I don't. 🙁 IMO it would be nice to add a note (linking to this issue, maybe) or remove the parts of the doc that don't work, at least.

mikem8361:
Fixed

dagood:
Nice, thanks! 😄

PR #84, opened 2019-01-15
Misc fixes and changes.

mikem8361:
Fixed Path.GetFileName on Windows paths running on Linux.

Fixed NugetSymbolServer project.

Fixed issue #83 and GitHub security alert about Microsoft.AspNetCore.Mvc.

changed dotnet-symbol version to 1.0.1.

Added default cache.

The same location as the SOS symbol server support.

$HOME/.dotnet/symbolcache on xplat.
%USERPROFILE%\.dotnet\symbolcache on Windows.

### DIFF
--- a/src/Microsoft.SymbolStore/KeyGenerators/ELFFileKeyGenerator.cs
+++ b/src/Microsoft.SymbolStore/KeyGenerators/ELFFileKeyGenerator.cs
@@ -87,7 +87,7 @@ namespace Microsoft.SymbolStore.KeyGenerators
                 }
                 else
                 {
-                    bool clrSpecialFile = s_coreClrSpecialFiles.Contains(Path.GetFileName(path));
+                    bool clrSpecialFile = s_coreClrSpecialFiles.Contains(GetFileName(path));
                     yield return BuildKey(path, IdentityPrefix, buildId, clrSpecialFile);
                 }
             }
@@ -104,7 +104,7 @@ namespace Microsoft.SymbolStore.KeyGenerators
                 if ((flags & KeyTypeFlags.ClrKeys) != 0)
                 {
                     /// Creates all the special CLR keys if the path is the coreclr module for this platform
-                    if (Path.GetFileName(path) == CoreClrFileName)
+                    if (GetFileName(path) == CoreClrFileName)
                     {
                         foreach (string specialFileName in s_coreClrSpecialFiles)
                         {
--- a/src/Microsoft.SymbolStore/KeyGenerators/KeyGenerator.cs
+++ b/src/Microsoft.SymbolStore/KeyGenerators/KeyGenerator.cs
@@ -88,9 +88,7 @@ namespace Microsoft.SymbolStore.KeyGenerators
         /// <returns>key</returns>
         protected static SymbolStoreKey BuildKey(string path, string id, bool clrSpecialFile = false)
         {
-            // The back slashes are changed to forward slashes because Path.GetFileName doesn't work on 
-            // Linux /MacOS if there are backslashes. Both back and forward slashes work on Windows.
-            string file = Uri.EscapeDataString(Path.GetFileName(path.Replace('\\', '/')).ToLowerInvariant());
+            string file = Uri.EscapeDataString(GetFileName(path).ToLowerInvariant());
             return BuildKey(path, null, id, file, clrSpecialFile);
         }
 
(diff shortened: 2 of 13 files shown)


tmat:
@mikem8361 Why should the default symbol cache be specific to .dotnet?
I would assume it would be shared across all frameworks and all debuggers on the machine.

We have recently discussed setting the default primary symbol cache to `%USERPROFILE%\.nuget\symbols`, since NuGet.org now supports symbol servers and we want it to be able to restore symbols to a local symbol cache (something like `nuget.exe restore -symbols` or `dotnet restore --symbols`).

@chuckries @anangaur @rrelyea

mikem8361:
Because this tool is dotnet and that is where the symbol download now in SOS caches. These tools have nothing to with nuget so that cache doesn't make any sense. $HOME/.dotnet/tool is where global tools like dotnet-symbol are installed.


-------- Original message --------

tmat:
I don't think it make sense for customers to have to manage one symbol cache for each tool they use. 
Besides, if the tools don't know of each other's cache then they will download the same symbol file multiple times wasting bandwidth and space on disk. 

Maybe `%USERPROFILE%\.nuget\symbols` is not the right place, maybe it's `%USERPROFILE%\.symbols`, but I believe we need a single directory to store symbols, not 3:  `%USERPROFILE%\AppData\Local\Temp\SymbolCache` (VS debugger), `%USERPROFILE%.dotnet\symbolcache` (dotnet-symbol),  `%USERPROFILE%\.nuget\symbols` (NuGet).

mikem8361:
You have a good point. windbg/cdb also have yet another cache usually in %ProgramData%\dbg\sym.

I would be fine with %USERPROFILE%\.symbols on Windows and $HOME/.symbols on xplat.

gregg-miskelly:
I don't think VS can change its default. So if you think it is useful to share the same location as VS, you should use %USERPROFILE%\AppData\Local\Temp\SymbolCache on Windows.

tmat:
@gregg-miskelly Why can't VS change the default?

mikem8361:
Ok, I think we should use the VS directory on Windows and my original xplat directory.  On xplat, I think it does make sense for VS, dotnet-symbol, SOS and nuget to put the ".dotnet" directory since these are dotnet or dotnet related symbols. I also think on xplat it should be lowercase:

Windows: %USERPROFILE%\AppData\Local\Temp\SymbolCache
Linux/MacOS: $HOME/.dotnet/symbolcache

@noahfalk, @tmat, @gregg-miskelly

mikem8361:
@gregg-miskelly does VS use the "TEMP" env var to get to this directory or does it explicitly %USERPROFILE%\AppData\Local\Temp\SymbolCache?

mikem8361:
@gregg-miskelly one more question: do you mind the directory being in $HOME/.dotnet for vsdbg? There isn't any native debugging for VSCode yet, but would it want to use a different path?

tmat:
Should't it be `symbol-cache` on linux? Not sure what the general convention is, but `symbolcache` is less readable.

anangaur:
We should all agree on one symbols folder - something that NuGet can also use in future while pre-fetching symbols (with an option) during restore or with any other command. 

What does dotnet-symbol do? Is it used for pre-caching symbols? For what?

gregg-miskelly:
>@gregg-miskelly does VS use the "TEMP" env var to get to this directory or does it explicitly %USERPROFILE%\AppData\Local\Temp\SymbolCache?

The VS code will call GetTempPath, and then append the "SymbolCache" directory to it. I would need to check the implementation, but I believe GetTempPath will always return the environment variable if it is present (which obviously it is 99.999% of the time).

> @gregg-miskelly one more question: do you mind the directory being in $HOME/.dotnet for vsdbg? There isn't any native debugging for VSCode yet, but would it want to use a different path?

To be clear: I assume you mean something like `$HOME/.dotnet/symbolcache` (or `$HOME/.dotnet/symbol-cache` linke Tomas's suggestion) and not directly in the '.dotnet' folder.

Personally I think it makes sense on Linux to have a shared symbol cache directory for the various .NET products, and if symbol server ever catches on with a non-.NET Linux audience through the VS/VS Code C++ then we can use a different cache directory. I don't think it makes sense to pick a more general purpose directory when may never have a solution in the C++ space.

mikem8361:
dotnet-symbol cracks core dumps, mini-dumps and other binaries and downloads the symbols for them from the Microsoft symbol server.

lldb doesn’t download symbols so this is to fill in the gap allowing all the .NET Core runtime symbols and binaries needed to debug a core dump be downloaded.

See https://github.com/dotnet/symstore/tree/master/src/dotnet-symbol#symbol-downloader-dotnet-cli-extension for more details.

Do you have any objections with “.dotnet” being in the symbol cache path? This also assumes you agree will the layout and indexing used in the cache by VS, dbgeng and these .dotnet tools. The specs for symbol server indexing are here: https://github.com/dotnet/symstore/tree/master/docs/specs#table-of-contents.

The cache layout is <file-name>/<index>/<file-name> all lowercased.

mikem8361:
Yes, I meant $HOME/.dotnet/symbolcache on Linux for vsdbg.

PR #85, opened 2019-01-23
Set RepositoryUrl/ProjectUrl to github repo.

mikem8361:
Needed to upload to nuget.

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -1,6 +1,7 @@
 <!-- Copyright (c)  Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information. -->
 <Project>
   <PropertyGroup>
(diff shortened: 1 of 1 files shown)


tmat:
This should be set by SourceLink package. Do you have SourceLink disabled? I remember there were some issues with SourceLink on some Linux distros, but those were fixed.

mikem8361:
RepositoryUrl was set to the internal mirror'ed DotNet-SymStore-Trusted repo.  This repo also still uses Roslyn-Toolset.

mikem8361:
I don't have SourceLink disabled by the way.

mikem8361:
When I tried to upload the latest dotnet-symbol tool package to nuget it complained about the ProjectUrl. That is where I saw that the Repo url was the internal one.

tmat:
I see. OK.

Issue #87, opened 2019-04-12
ERROR: The operation was canceled. -> Unable to read data from the transport connection: Operation canceled.


maxd:
Hello,

When I run `dotnet symbol` to download debug symbols for .NET Core SDK 2.1.10 it throw the following error:

```
...
Writing: /usr/share/dotnet/shared/Microsoft.NETCore.App/2.1.10/System.Diagnostics.FileVersionInfo.pdb
Cached: /root/.dotnet/symbolcache/system.componentmodel.annotations.pdb/5b25c24677e342c39b52d7d0bb347779FFFFFFFF/system.componentmodel.annotations.pdb
Writing: /usr/share/dotnet/shared/Microsoft.NETCore.App/2.1.10/System.ComponentModel.Annotations.pdb
Cached: /root/.dotnet/symbolcache/system.private.corelib.pdb/d86a24c58e91402d8f6d21028185ef75FFFFFFFF/system.private.corelib.pdb
Writing: /usr/share/dotnet/shared/Microsoft.NETCore.App/2.1.10/System.Private.CoreLib.pdb
ERROR: The operation was canceled. -> Unable to read data from the transport connection: Operation canceled.
```

Looks like `dotnet symbol` is trying to download large file and can't do it because exceed read timeout. Could you please check it?

Steps to reproduce:

1. Create `Dockerfile`:

    ```
    FROM mcr.microsoft.com/dotnet/core/sdk:2.1.603@sha256:8eda326c4c0bd038f3cc5aeb5c8af8374b88f1b90eb444ed91bab42f2e17840f

    RUN dotnet tool install -g dotnet-symbol

    ENV PATH=$PATH:/root/.dotnet/tools

    RUN dotnet symbol --diagnostics --symbols --output /usr/share/dotnet/shared/Microsoft.NETCore.App/2.1.10 '/usr/share/dotnet/shared/Microsoft.NETCore.App/2.1.10/*'
    ```

1. Run `docker build` to build this docker file:

    ```
    docker build . -f Dockerfile
    ```

It should show error described above.

Thank you.

mikem8361:
I just got back from vacation. I'll look into this as soon as I can.

mikem8361:
Of course it doesn't repro for me on my machine/network.  

[...]

PR #88, opened 2019-04-15
Various dotnet-symbol and symstore fixes

mikem8361:
This is an accumulation of dotnet-symbol/symstore fixes.

Fix reads across contiguous adjacent segments. This is the same problem and changes applied to the CLRMD's ELF dump reader recently.

Catch exceptions around create cache directory. This catches and skip caching when the cached file was written previously as superuser.  

Increase the request timeout from 100 seconds (1:40min) to 4 minutes. Issue #87.

### DIFF
--- a/src/Microsoft.SymbolStore/SymbolStores/CacheSymbolStore.cs
+++ b/src/Microsoft.SymbolStore/SymbolStores/CacheSymbolStore.cs
@@ -15,7 +15,6 @@ namespace Microsoft.SymbolStore.SymbolStores
             : base(tracer, backingStore)
         {
             CacheDirectory = cacheDirectory ?? throw new ArgumentNullException(nameof(cacheDirectory));
-            Directory.CreateDirectory(cacheDirectory);
         }
 
         protected override Task<SymbolStoreFile> GetFileInner(SymbolStoreKey key, CancellationToken token)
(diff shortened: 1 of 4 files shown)


src/Microsoft.SymbolStore/SymbolStores/CacheSymbolStore.cs
```diff
@@ -15,7 +15,13 @@ public CacheSymbolStore(ITracer tracer, SymbolStore backingStore, string cacheDi
             : base(tracer, backingStore)
         {
             CacheDirectory = cacheDirectory ?? throw new ArgumentNullException(nameof(cacheDirectory));
-            Directory.CreateDirectory(cacheDirectory);
+            try
+            {
+                Directory.CreateDirectory(cacheDirectory);
```
noahfalk: Could we skip trying to create the directory entirely at this point in the code? It appears the code below will lazily create it when we download the first file. If the directory needs to exist eagerly then we probably shouldn't ignore the exceptions that occur when creation fails.

src/Microsoft.SymbolStore/SymbolStores/HttpSymbolStore.cs
```diff
@@ -41,16 +41,21 @@ public HttpSymbolStore(ITracer tracer, SymbolStore backingStore, Uri symbolServe
             {
                 throw new ArgumentException(nameof(symbolServerUri));
             }
+
             // Normal unauthenticated client
-            _client = new HttpClient();
+            _client = new HttpClient {
+                Timeout = TimeSpan.FromMinutes(4)
```
noahfalk: Can you give some background on this? Naively it seems a little odd that we would specify a non-default timeout, or that we would pick a timeout that was so long.

mikem8361: The issue in the PR description (#87) has more details from the customer. In his case, the default timeout (100 secs) timed out on a long file download on his machine in a docker container.  I basically doubled the timeout and it fixed his problem. I don't have a sense of any downside, but the docs on this property don't give a lot of detail.

src/Microsoft.SymbolStore/SymbolStores/CacheSymbolStore.cs
mikem8361: I don't remember if there were any reasons to eagerly create the base cache directory and your right that the below CreateDirectory should create all the directories in the path.  

I'll remove it from the constructor and do some testing.

src/Microsoft.SymbolStore/SymbolStores/HttpSymbolStore.cs
noahfalk: My only worry would be if downloading symbols from an unresponsive server became really painful, for example downloading 50 files where each of them waits 4 minutes and then gives up. Given that this is only double the previous I'll assume if this was an issue people would have already complained. I'm fine leaving this as-is and then in the future if someone complains we'll think harder about how to address it.

noahfalk: (and whoops, your description had reasonable info and I just missed it! My bad)

PR #94, opened 2019-04-24
Fix dotnet-symbol versioning/updating to 1.0.3

mikem8361:


tmat:
FYI: https://github.com/dotnet/arcade/pull/2615 will allow us to auto-generate the package versions for global tools, so we don't need to bump them manually.

mikem8361:
That would be great.  Finding time to update the symstore repo with the latest arcade after this PR is in might be a problem.

We also need this in our diagnostics repo, too.

PR #97, opened 2019-05-31
Use DBI stream header Age instead of PDB header Age

mikem8361:


src/Microsoft.FileFormats/PDB/PDBFile.cs
```diff
@@ -14,19 +14,22 @@ public class PDBFile
...
             _nameStream = new Lazy<PDBNameStream>(() => new PDBNameStream(Streams[1]));
+            _dbiStream = new Lazy<DbiStream>(() => new DbiStream(Streams[3]));
         }
 
         public PDBFileHeader Header { get { return _header.Value; } }
         public IList<Reader> Streams { get { return _streams.Value; } }
         public PDBNameStream NameStream { get { return _nameStream.Value; } }
-        public uint Age { get { return NameStream.Header.Age; } }
+        public DbiStream DbiStream { get { return _dbiStream.Value; } }
+        public uint Age { get { return DbiStream.Header.Age; } }
```
noahfalk: Given that there are two Ages in the file, maybe change the name to DbiAge so its clear which one is being refered to?

mikem8361: I added the DbiAge property like you suggested.

PR #103, opened 2019-07-02
Add SymbolChecksum header to HttpRequest

### LINKED ISSUE #102: SymClient returns 403 Forbidden when downloading symbols from NuGet
image

`SymbolChecksum` needs implementing.

image

Working on PR.

nosami:
Fixes #102

image

### DIFF
--- a/src/Microsoft.FileFormats/PE/PEStructures.cs
+++ b/src/Microsoft.FileFormats/PE/PEStructures.cs
@@ -332,7 +332,7 @@ namespace Microsoft.FileFormats.PE
         /// </summary>
         /// <param name="bytes">array of bytes</param>
         /// <returns>hex string</returns>
-        private static string ToHexString(byte[] bytes)
+        public static string ToHexString(byte[] bytes)
         {
             if (bytes == null)
             {
--- a/src/Microsoft.FileFormats/PE/PEStructures.cs
+++ b/src/Microsoft.FileFormats/PE/PEStructures.cs
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
 
 using System;
+using System.Linq;
 
 namespace Microsoft.FileFormats.PE
 {
@@ -308,11 +309,11 @@ namespace Microsoft.FileFormats.PE
     #endregion
 
     /// <summary>
-    /// VS_PDBCHECKSUM struct
+    /// Pair of a checksum algorithm name (ex: "SHA256") and the bytes of the checksum.
     /// </summary>
-    public class VsPdbChecksum : TStruct
+    public class PdbChecksum : TStruct
     {
-        public VsPdbChecksum(string algorithmName, byte[] checksum)
+        public PdbChecksum(string algorithmName, byte[] checksum)
         {
             AlgorithmName = algorithmName;
             Checksum = checksum;
@@ -320,5 +321,24 @@ namespace Microsoft.FileFormats.PE
 
         public string AlgorithmName { get; }
         public byte[] Checksum { get; }
+
+        public override string ToString()
+        {
+            return $"{AlgorithmName}:{ToHexString(Checksum)}";
+        }
+
+        /// <summary>
+        /// Convert an array of bytes to a lower case hex string.
+        /// </summary>
+        /// <param name="bytes">array of bytes</param>
+        /// <returns>hex string</returns>
+        private static string ToHexString(byte[] bytes)
+        {
+            if (bytes == null)
+            {
+                throw new ArgumentNullException(nameof(bytes));
+            }
+            return string.Concat(bytes.Select(b => b.ToString("x2")));
+        }
     }
 }
--- a/src/Microsoft.FileFormats/PE/PEStructures.cs
+++ b/src/Microsoft.FileFormats/PE/PEStructures.cs
@@ -184,6 +184,7 @@ namespace Microsoft.FileFormats.PE
         Bbt = 10,
         Reproducible = 16,
         EmbeddedPortablePdb = 17,
+        PdbChecksum = 19
     };
 
     /// <summary>
@@ -305,4 +306,19 @@ namespace Microsoft.FileFormats.PE
     }
 
     #endregion
+
+    /// <summary>
+    /// VS_PDBCHECKSUM struct
+    /// </summary>
+    public class VsPdbChecksum : TStruct
+    {
+        public VsPdbChecksum(string algorithmName, byte[] checksum)
+        {
+            AlgorithmName = algorithmName;
+            Checksum = checksum;
+        }
+
+        public string AlgorithmName { get; }
+        public byte[] Checksum { get; }
+    }
 }
--- a/src/Microsoft.SymbolStore/ChecksumValidator.cs
+++ b/src/Microsoft.SymbolStore/ChecksumValidator.cs
@@ -1,0 +1,140 @@
+﻿using Microsoft.FileFormats.PE;
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Security.Cryptography;
+using System.Text;
+
+namespace Microsoft.SymbolStore
+{
+    public class InvalidChecksumException : Exception
+    {
+        public InvalidChecksumException(string message) : base(message)
+        {
+
+        } 
+    }
+
+    class ChecksumValidator
+    {
+        private const string pdbStreamName = "#Pdb";
+        private const uint pdbIdSize = 20;
+        private readonly ITracer _tracer;
+
+        public ChecksumValidator(ITracer tracer)
+        {
+            _tracer = tracer;
+        }
+
+        internal void Validate(Stream pdbStream, IEnumerable<PdbChecksum> pdbChecksums)
+        {
+            using (MemoryStream ms = new MemoryStream())
+            {
+                pdbStream.CopyTo(ms);
+                byte[] bytes = ms.ToArray();
+                uint offset = 0;
+
+                try
+                {
+                    offset = GetPdbStreamOffset(pdbStream);
+                }
+                catch (Exception ex)
+                {
+                    _tracer.Error(ex.Message);
+                    throw;
+                }
+
+                for (int i = 0; i <= pdbIdSize; i++)
+                {
+                    bytes[i + offset] = 0;
+                }
+
+                foreach (var checksum in pdbChecksums)
+                {
+                    _tracer.Information($"Testing checksum: {checksum}");
+
+                    var algorithm = HashAlgorithm.Create(checksum.AlgorithmName);
+                    if (algorithm != null)
+                    {
+                        var hash = algorithm.ComputeHash(bytes);
+                        if (hash.SequenceEqual(checksum.Checksum))
+                        {
+                            // If any of the checksums are OK, we're good
+                            _tracer.Information($"Found checksum match {checksum}");
+                            return;
+                        }
+                    }
+                    else
+                    {
+                        throw new ArgumentException("Unknown hash algorithm");
+                    }
+                }
+            }
+            throw new InvalidChecksumException("PDB checksum mismatch");
+        }
+
+        private static uint GetPdbStreamOffset(Stream pdbStream)
+        {
+            pdbStream.Position = 0;
+            var reader = new BinaryReader(pdbStream);
+            {
+                uint signature = reader.ReadUInt32();
+                pdbStream.Seek(4 + // Signature
+                               2 + // Version Major
+                               2 + // Version Minor
+                               4,  // Reserved)
+                               SeekOrigin.Begin);
+
+                // skip the version string
+                uint versionStringSize = 0;
+                versionStringSize = reader.ReadUInt32();
+
+                pdbStream.Seek(versionStringSize, SeekOrigin.Current);
+
+                // storage header
+                pdbStream.Seek(2, SeekOrigin.Current);
+
+                // read the stream headers
+                ushort streamCount = reader.ReadUInt16();
+                uint streamOffset;
+                string streamName;
+
+                for (int i = 0; i < streamCount; i++)
+                {
+                    var pos = pdbStream.Position;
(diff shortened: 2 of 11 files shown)


src/Microsoft.FileFormats/PE/PEStructures.cs
```diff
@@ -305,4 +306,19 @@ public class VsVersionInfo  : TStruct
     }
 
     #endregion
+
+    /// <summary>
+    /// VS_PDBCHECKSUM struct
```
gregg-miskelly: How about: `Pair of a checksum algorithm name (ex: "SHA256") and the bytes of the checksum.`

src/Microsoft.SymbolStore/SymbolStoreKey.cs
```diff
@@ -32,18 +32,30 @@ public sealed class SymbolStoreKey
         /// </summary>
         public static SymbolStoreKey[] EmptyArray = new SymbolStoreKey[0];
 
+        /// <summary>
```
gregg-miskelly: Don't you need this to be a collection of `VsPdbChecksum` instead -- a PDB can have multiple checksums.

src/Microsoft.SymbolStore/KeyGenerators/KeyGenerator.cs
```diff
@@ -85,11 +86,12 @@ public virtual bool IsDump()
         /// <param name="path">full path of file or binary</param>
         /// <param name="id">id string</param>
         /// <param name="clrSpecialFile">if true, the file is one the clr special files</param>
+        /// <param name="checksum">Checksum of pdb file. May be null.</param>
         /// <returns>key</returns>
-        protected static SymbolStoreKey BuildKey(string path, string id, bool clrSpecialFile = false)
+        protected static SymbolStoreKey BuildKey(string path, string id, bool clrSpecialFile = false, VsPdbChecksum checksum = null)
```
gregg-miskelly: I believe all of these should be `IEnumerable<VsPdbChecksum>`

src/Microsoft.SymbolStore/SymbolStores/HttpSymbolStore.cs
```diff
@@ -75,6 +75,14 @@ public void ResetClientFailure()
         protected override async Task<SymbolStoreFile> GetFileInner(SymbolStoreKey key, CancellationToken token)
         {
             Uri uri = GetRequestUri(key.Index);
+
+            if (key.ChecksumAlgorithmName != null && key.ChecksumAsHex != null)
```
gregg-miskelly: Sorry if this is obvious, but I think we need validation support before we complete this PR. We could either build it directly into this library, or provide a call back if we want to avoid possible version conflicts with the portable PDB reader assembly.

src/Microsoft.FileFormats/PE/PEStructures.cs
```diff
@@ -305,4 +306,19 @@ public class VsVersionInfo  : TStruct
...
+    /// </summary>
+    public class VsPdbChecksum : TStruct
```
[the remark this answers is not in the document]
mikem8361: I agree.

src/Microsoft.SymbolStore/KeyGenerators/KeyGenerator.cs
mikem8361: These helper functions build just one key at a time so I not sure how they would an enumeration of checksums.

src/Microsoft.SymbolStore/KeyGenerators/PortablePDBFileKeyGenerator.cs
```diff
@@ -76,11 +77,11 @@ public override IEnumerable<SymbolStoreKey> GetKeys(KeyTypeFlags flags)
         /// <param name="path">file name and path</param>
         /// <param name="pdbId">pdb guid</param>
         /// <returns>symbol store key</returns>
-        public static SymbolStoreKey GetKey(string path, Guid pdbId)
+        public static SymbolStoreKey GetKey(string path, Guid pdbId, VsPdbChecksum checksum)
```
mikem8361: You can't change this public function's signature.  Need to add another function with the checksum parameter.  Giving checksum a default value won't work either.

mikem8361: Never mind about the default value comment. Setting checksum = null should work.

src/Microsoft.SymbolStore/SymbolStores/HttpSymbolStore.cs
mikem8361: @gregg-miskelly , I don't understand what you mean by "validation support".

gregg-miskelly: This code should verify that the downloaded PDB has a hash that matches one of the hashes in the hash collections.

mikem8361:
Yes, that makes sense.

src/Microsoft.SymbolStore/KeyGenerators/KeyGenerator.cs
gregg-miskelly: A PE file can have more than one checksum, if it does, a result is valid as long as at least one match is correct.

This is for two different scenarios -
1. New hash algorithm: as hash algorithms are deprecated, we may want to list more than one has value to maintain compatibly with old systems while supporting strong cypto.
2. Alternative PDBs: The same binary may have multiple PDB files that are all valid. Today none of these scenarios make sense for portable PDBs, but maybe sometime they will. Scenarios for native Windows PDBs -- you can have (1) a normal PDB (2) a PDB with srcsrv stream injected -and- (3) a "public" PDB (no types or source lines).

mikem8361: I think I understand that PE's can have more than one checksum, but the enumeration/validation would have to be done at the higher level in the PEFileKeyGenerator.GetKeys implementation like Jason did unless I misunderstood your point.

gregg-miskelly: No, what Jason did in `PEFileKeyGenerator` was incorrect. What it should have done was gather up all the checksums and assigned them to all the keys.

src/Microsoft.SymbolStore/SymbolStores/HttpSymbolStore.cs
nosami: I wondered about whether or not the validation support belonged in here. Thanks for the clarification.

gregg-miskelly:
Aside from needing to add validation or a validation call back, these changes look good to me.

mikem8361:
LGTM.

nosami:
Yeah... I will add the validation, I just got sidetracked with other stuff. I will see how easy it would be to add it in here first.

nosami:
OK. I think this is ready for another 👀

src/Microsoft.SymbolStore/SymbolStores/HttpSymbolStore.cs
```diff
@@ -75,9 +79,26 @@ public void ResetClientFailure()
...
+                Tracer.Information($"SymbolChecksum: {checksumHeader}");
+                _client.DefaultRequestHeaders.Add("SymbolChecksum", checksumHeader);
+            }
+
+            stream = await GetFileStream(uri, token);
             if (stream != null)
             {
+                if(needsChecksumMatch)
+                {
+                    var validator = new ChecksumValidator(_tracer);
```
nosami: maybe I should just make this static

mikem8361: Sounds good.  Minor nit: there needs to be a space after the if (needsChecksumMatch)

src/Microsoft.SymbolStore/ChecksumValidator.cs
```diff
@@ -0,0 +1,125 @@
...
+    {
+        private const string pdbStreamName = "#Pdb";
+        private const uint pdbIdSize = 20;
+
+        internal static void Validate(ITracer tracer, Stream pdbStream, IEnumerable<PdbChecksum> pdbChecksums)
+        {
+            using (MemoryStream ms = new MemoryStream())
+            {
+                pdbStream.CopyTo(ms);
+                byte[] bytes = ms.ToArray();
```
gregg-miskelly: I believe this method of copying the stream will result in two different in-memory copyies of all the bytes. I think you want:

```
        byte[] bytes = new byte[pdbStream.Length];
        if (pdbStream.Read(bytes, offset: 0, count: bytes.Length) != bytes.Length)
        {
            throw ...
        }
```

src/Microsoft.SymbolStore/ChecksumValidator.cs
```diff
@@ -0,0 +1,125 @@
...
+                    }
+                }
+            }
+            throw new InvalidChecksumException("PDB checksum mismatch");
+        }
+
+        private static uint GetPdbStreamOffset(Stream pdbStream)
+        {
+            pdbStream.Position = 0;
+            var reader = new BinaryReader(pdbStream);
```
gregg-miskelly: I believe this will close the pdbStream when the BinaryReader is finialized. Do you want the overload with the `leaveOpen` argument?

src/Microsoft.SymbolStore/ChecksumValidator.cs
```diff
@@ -0,0 +1,125 @@
...
+            {
+                uint signature = reader.ReadUInt32();
```
gregg-miskelly: Do you want to check the signature?

src/Microsoft.SymbolStore/ChecksumValidator.cs
```diff
@@ -0,0 +1,125 @@
...
+                    {
+                        var hash = algorithm.ComputeHash(bytes);
+                        if (hash.SequenceEqual(checksum.Checksum))
+                        {
+                            // If any of the checksums are OK, we're good
+                            tracer.Information($"Found checksum match {checksum}");
+                            return;
+                        }
+                    }
+                    else
```
gregg-miskelly: You don't want to throw in this case unless all the algorithm names are unknown -- this might be a new checksum that this framework doesn't support, but a newer framework does.

gregg-miskelly: Also, it should probably still throw an InvalidChecksumException, but with a different message.

src/Microsoft.SymbolStore/ChecksumValidator.cs
```diff
@@ -0,0 +1,125 @@
...
+                    {
+                        throw new ArgumentException("Unknown hash algorithm");
+                    }
+                }
+            }
+            throw new InvalidChecksumException("PDB checksum mismatch");
```
gregg-miskelly: To make sure I understand, the idea here is that the InvalidChecksumException makes it all the way back to the client, is that right? If so, might we worth calling that out in the doc comments for the public API [here](https://github.com/dotnet/symstore/blob/18f84ad901fbf454dccc846be2f0cadf1a53e25a/src/Microsoft.SymbolStore/SymbolStores/SymbolStore.cs#L39)

src/Microsoft.SymbolStore/ChecksumValidator.cs
gregg-miskelly: Dispose the reader?

src/Microsoft.SymbolStore/ChecksumValidator.cs
nosami: looks better, thanks

src/Microsoft.SymbolStore/ChecksumValidator.cs
nosami: TIL

src/Microsoft.SymbolStore/ChecksumValidator.cs
nosami: No. Removed.

nosami:
Addressed feedback.

src/Microsoft.SymbolStore/ChecksumValidator.cs
nosami: Just been testing this and I think my code was depending on there being two different copies (one with zeroed bytes for the checksum and the unaltered stream for writing to disk)

I'm getting `System.BadImageFormatException: Invalid COR20 header signature.` when I try and load the pdb and I think this is probably the reason why.

nosami: as an aside - why do we need to zero the signature bytes for the checksum? Are they not deterministic?

nosami: Reverted this change and it works fine.

nosami:
I put the stream copy code back in place as it was needed. One stream has the zeroed bytes for the checksum and the other is the unaltered bytes ready to be written to disk.

I guess alternatively, I could put the signature bytes back into the byte array before writing to disk?

nosami:
OK, it works after the last commit :)

src/Microsoft.SymbolStore/ChecksumValidator.cs
nosami: Fixed by making a backup of the pdb Id and restoring before writing to disk to save making a copy of the whole stream.

gregg-miskelly:
I am very confused why `Stream.Read` decided to change the underlying stream to hold a reference to `bytes` rather than just copying its data out to `bytes`. But whatever, your fix certainly works.

mikem8361:
Are we ready to merge this?

nosami:
This PR is working for me in VSMac.

mikem8361:
The new version of the Microsoft.SymbolStore package 1.0.0-dev-64311-01 on https://dotnet.myget.org/F/symstore/api/v2 feed.

I'll try to get a new version of the dotnet-symbol global tool up on nuget but it may be a bit.

mikem8361:
The new dotnet-symbol has been published to nuget. Either do:

    dotnet tool update -g dotnet-symbol

or

    dotnet tool install -g dotnet-symbol 

if the first time.

nosami:
Thanks everyone for helping me get this across the line.

PR #107, opened 2019-08-08
Restore stream position after checksum validation

gregg-miskelly:
At least on CoreCLR using the symbol store's caching reader, the stream position needs to be restored after checksum validation or it will be omitted from the file. This adds code to do this.


gregg-miskelly:
CC @mikem8361 @nosami

nosami:
Thanks!

gregg-miskelly:
@mikem8361 could you generate a new package to the feed with this change?

mikem8361:
The version of the symstore packages (Microsoft.SymbolStore, Microsoft.FileFormats) with this fix is 1.0.0-dev-64412-01.  They are on this myget feed:  https://dotnet.myget.org/F/symstore/api/v2

mikem

nosami:
Cheers @mikem8361  !

Issue #130, opened 2019-10-10
Unreasonably large ELFHeader when generating symbols from coredump


taion809:
Hello,
We've created a coredump using the createdump utility.  Upon trying to use dotnet-dump analyze we see the following:
```
> clrstack -all
...
00007F85D2FFA4D0 00007F85FE7F08B0 /app/NLog.dll!Unknown
00007F85D2FFA500 00007F85FCF5B2AF /app/NLog.dll!Unknown
00007F85D2FFA570 00007F85FCF55E96 /app/NLog.dll!Unknown
00007F85D2FFA5B0 00007F85FCF55DC2 /app/NLog.dll!Unknown
00007F85D2FFA5F0 00007F85FCF559C5 /app/NLog.dll!Unknown
...
```

Following [this documentation](https://github.com/dotnet/diagnostics/blob/master/documentation/debugging-coredump.md#getting-symbols), when attempting to generate symbols from the dump we see the following
```
> dotnet-symbol ./3f13aaf3-3485-01c7-1ee6-82a583e97eb2.coredump
Downloading from http://msdl.microsoft.com/download/symbols/
ERROR: Unreasonably large ELFHeader ProgramHeaderCount
```

How should we proceed with debugging this?

dotnet info:
```
.NET Core SDK (reflecting any global.json):
 Version:   3.0.100
 Commit:    04339c3a26

Runtime Environment:
 OS Name:     ubuntu
 OS Version:  18.04
 OS Platform: Linux
 RID:         ubuntu.18.04-x64
 Base Path:   /usr/share/dotnet/sdk/3.0.100/

Host (useful for support):
  Version: 3.0.0
  Commit:  95a0a61858

.NET Core SDKs installed:
  2.2.402 [/usr/share/dotnet/sdk]
  3.0.100 [/usr/share/dotnet/sdk]

.NET Core runtimes installed:
  Microsoft.AspNetCore.All 2.2.7 [/usr/share/dotnet/shared/Microsoft.AspNetCore.All]
  Microsoft.AspNetCore.App 2.2.7 [/usr/share/dotnet/shared/Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 3.0.0 [/usr/share/dotnet/shared/Microsoft.AspNetCore.App]
  Microsoft.NETCore.App 2.2.7 [/usr/share/dotnet/shared/Microsoft.NETCore.App]
[...]

Issue #117, opened 2019-09-14
Fix SymbolStores that don't return SymbolStoreFile.Stream position to 0


mikem8361:
It may be only the newer DirectorySymbolStore, but the other ones need to be checked.

The DirectorySymbolStore can't find the DAC or DBI in a directory for the special coreclr based index.

Add info trace when the DirectorySymbolStore finds a file.

Add "SymbolStore.Equals" overrides to all the symbol stores.

PR #131, opened 2019-10-15
Fix various issues

mikem8361:
Fix some symbol publishing problems.

Fix issue #130 by increasing the section/program header count to 30000.

Fix issue #117. Always set the symbol file stream position to 0.

Add Equals, GetHashCode and ToString to the SymbolStore implementations

Fix issue #106.

### DIFF
--- a/src/Microsoft.FileFormats/ELF/ELFStructures.cs
+++ b/src/Microsoft.FileFormats/ELF/ELFStructures.cs
@@ -138,7 +138,7 @@ namespace Microsoft.FileFormats.ELF
         {
             get
             {
-                return new ValidationRule("Unreasonably large ELFHeader ProgramHeaderCount", () => ProgramHeaderCount <= 10000);
+                return new ValidationRule("Unreasonably large ELFHeader ProgramHeaderCount", () => ProgramHeaderCount <= 30000);
             }
         }
 
@@ -146,7 +146,7 @@ namespace Microsoft.FileFormats.ELF
         {
             get
             {
-                return new ValidationRule("Unreasonably large ELFHeader SectionHeaderCount", () => SectionHeaderCount <= 10000);
+                return new ValidationRule("Unreasonably large ELFHeader SectionHeaderCount", () => SectionHeaderCount <= 30000);
             }
         }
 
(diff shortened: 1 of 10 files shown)


src/Microsoft.FileFormats/ELF/ELFStructures.cs
```diff
@@ -138,15 +138,15 @@ public ValidationRule IsProgramHeaderCountReasonable
...
-                return new ValidationRule("Unreasonably large ELFHeader ProgramHeaderCount", () => ProgramHeaderCount <= 10000);
+                return new ValidationRule("Unreasonably large ELFHeader ProgramHeaderCount", () => ProgramHeaderCount <= 30000);
             }
         }
 
         public ValidationRule IsSectionHeaderCountReasonable
         {
             get
             {
-                return new ValidationRule("Unreasonably large ELFHeader SectionHeaderCount", () => SectionHeaderCount <= 10000);
```
hoyosjs: What generates so many headers?

src/dotnet-symbol/README.md
```diff
@@ -52,6 +52,8 @@ To verify a symbol package on a local VSTS symbol server:
 
 ## Notes ##
 
+Symbol download is only supported for official .NET Core runtime versions installed from our official [web site](https://dotnet.microsoft.com/download/dotnet-core). Runtimes from community sites like [archlinux](https://www.archlinux.org/packages/community/x86_64/dotnet-runtime/) are not support.
```
hoyosjs:
```suggestion
Symbol download is only supported for official .NET Core runtime versions acquired through official channels such as [the official web site](https://dotnet.microsoft.com/download/dotnet-core) and the [default sources in the dotnet installation scripts](https://docs.microsoft.com/en-us/dotnet/core/tools/dotnet-install-scripts). Runtimes obtained from community sites like [archlinux](https://www.archlinux.org/packages/community/x86_64/dotnet-runtime/) are not supported. 
```

src/Microsoft.FileFormats/ELF/ELFStructures.cs
mikem8361: `createdump` generates them. `createdump` is also going to be changed not to generate all the
"not backed by memory" program header notes which will reduce the number of program headers significantly.

src/dotnet-symbol/README.md
mikem8361: Done.

PR #135, opened 2019-10-25
Host program support

mikem8361:
Added the --host-only option that downloads the host program usually "dotnet"
or "apphost" renamed to the program/project name for self-contained apps.

### DIFF
--- a/src/Microsoft.FileFormats/PE/PEStructures.cs
+++ b/src/Microsoft.FileFormats/PE/PEStructures.cs
@@ -105,11 +105,11 @@ namespace Microsoft.FileFormats.PE
     [Flags]
     public enum ImageFile : ushort
     {
-        RELOCS_STRIPPED         = 0x0001,
-        EXECUTABLE_IMAGE        = 0x0002,
-        LARGE_ADDRESS_AWARE     = 0x0020,
-        SYSTEM                  = 0x1000,
-        DLL                     = 0x2000,
+        RelocsStripped      = 0x0001,
+        ExecutableImage     = 0x0002,
+        LargeAddressAware   = 0x0020,
+        System              = 0x1000,
+        Dll                 = 0x2000,
     }
 
     /// <summary>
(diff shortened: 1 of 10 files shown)


mikem8361:
Ping, just in case you miss this.

src/Microsoft.FileFormats/PE/PEStructures.cs
```diff
@@ -99,6 +99,19 @@ public enum ImageFileMachine
         Arm64       = 0xAA64
     }
 
+    /// <summary>
+    /// Characteristics (IMAGE_FILE)
+    /// </summary>
+    [Flags]
+    public enum ImageFile : ushort
+    {
+        RELOCS_STRIPPED         = 0x0001,
```
noahfalk: Nit: We've typically used Pascal casing following BCL coding style rules

src/Microsoft.SymbolStore/KeyGenerators/KeyGenerator.cs
```diff
@@ -39,6 +41,11 @@ public enum KeyTypeFlags
         /// the Portable and Windows PDBs are available on the symbol server.
         /// </summary>
         ForceWindowsPdbs = 0x08,
+
+        /// <summary>
+        /// Generate keys for the host program.
```
noahfalk: Can we add a little more description about what the 'host program' relationship is? I assume for a binary 'foo', it is a key with the same buildid renamed to 'apphost'?

src/Microsoft.FileFormats/PE/PEStructures.cs
mikem8361: I was being lazy and copied this enum from clrmd. I'll fix it.

src/Microsoft.SymbolStore/KeyGenerators/KeyGenerator.cs
mikem8361: Done.

PR #170, opened 2020-03-06
Fix URI double escaping problem

mikem8361:
This change moves the URI escaping from where we build the indexes (KeyGenerator.cs) to the http symbol store.  This is so the symbol uploader uses the un-escaped index and the downloader escapes the URI before it gets sent on the network.

The Edge Engineering team uses the SymUploader utility to upload symbol files with blanks and windbg couldn't download them.

### DIFF
--- a/src/Microsoft.SymbolStore/KeyGenerators/KeyGenerator.cs
+++ b/src/Microsoft.SymbolStore/KeyGenerators/KeyGenerator.cs
@@ -99,7 +99,7 @@ namespace Microsoft.SymbolStore.KeyGenerators
         /// <returns>key</returns>
         protected static SymbolStoreKey BuildKey(string path, string id, bool clrSpecialFile = false, IEnumerable<PdbChecksum> pdbChecksums = null)
         {
-            string file = Uri.EscapeDataString(GetFileName(path).ToLowerInvariant());
+            string file = GetFileName(path).ToLowerInvariant();
             return BuildKey(path, null, id, file, clrSpecialFile, pdbChecksums);
         }
 
@@ -114,7 +114,7 @@ namespace Microsoft.SymbolStore.KeyGenerators
         /// <returns>key</returns>
         protected static SymbolStoreKey BuildKey(string path, string prefix, byte[] id, bool clrSpecialFile = false, IEnumerable<PdbChecksum> pdbChecksums = null)
         {
-            string file = Uri.EscapeDataString(GetFileName(path).ToLowerInvariant());
+            string file = GetFileName(path).ToLowerInvariant();
             return BuildKey(path, prefix, id, file, clrSpecialFile, pdbChecksums);
         }
 
--- a/src/Microsoft.SymbolStore/SymbolStoreKey.cs
+++ b/src/Microsoft.SymbolStore/SymbolStoreKey.cs
@@ -4,6 +4,7 @@ using Microsoft.FileFormats.PE;
 using System;
 using System.Collections.Generic;
 using System.Diagnostics;
+using System.IO;
 using System.Linq;
 
 namespace Microsoft.SymbolStore
@@ -82,6 +83,8 @@ namespace Microsoft.SymbolStore
             return string.Equals(Index, right.Index);
         }
 
+        private static HashSet<char> s_invalidChars = new HashSet<char>(Path.GetInvalidFileNameChars());
+
         /// <summary>
         /// Validates a symbol index.
         ///
@@ -104,7 +107,7 @@ namespace Microsoft.SymbolStore
                     if (char.IsLetterOrDigit(c)) {
                         continue;
                     }
-                    if (c == '_' || c == '-' || c == '.' || c == '%') {
+                    if (!s_invalidChars.Contains(c)) {
                         continue;
                     }
                     return false;
--- a/src/Microsoft.SymbolStore/SymbolStores/HttpSymbolStore.cs
+++ b/src/Microsoft.SymbolStore/SymbolStores/HttpSymbolStore.cs
@@ -121,7 +121,7 @@ namespace Microsoft.SymbolStore.SymbolStores
 
         protected Uri GetRequestUri(string index)
         {
-            if (!Uri.TryCreate(Uri, Uri.EscapeUriString(index), out Uri requestUri))
+            if (!Uri.TryCreate(Uri, Uri.EscapeDataString(index), out Uri requestUri))
             {
                 throw new ArgumentException(nameof(index));
             }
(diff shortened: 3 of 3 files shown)


src/Microsoft.SymbolStore/SymbolStores/HttpSymbolStore.cs
```diff
@@ -121,7 +121,7 @@ protected override async Task<SymbolStoreFile> GetFileInner(SymbolStoreKey key,
 
         protected Uri GetRequestUri(string index)
         {
-            if (!Uri.TryCreate(Uri, index, out Uri requestUri))
```
hoyosjs: Are we sure `EscapeUriString` is the right API? This closes the number of unsupported files, as `EscapeDataString` would always escape `:/?#[]@!$&'()*+,;=`. Hence, this covers the space case, but on several ext file systems most of these are valid in the file names. These would need to be escaped on each part of the triplet I believe. However, this is such an obscure case that I'm ok with this part of the change.

src/Microsoft.SymbolStore/KeyGenerators/KeyGenerator.cs
```diff
@@ -99,7 +99,7 @@ public virtual bool IsDump()
         /// <returns>key</returns>
         protected static SymbolStoreKey BuildKey(string path, string id, bool clrSpecialFile = false, IEnumerable<PdbChecksum> pdbChecksums = null)
         {
-            string file = Uri.EscapeDataString(GetFileName(path).ToLowerInvariant());
+            string file = GetFileName(path).ToLowerInvariant();
```
hoyosjs: Doing this changes the key for file names. Any file that had these characters will have the key changed on windows: `#[]@!$&'()+,;=`. The parenthesis is the only one that I've seen used in real life. They used to be escaped and now the URL won't escape them. Some of which are invalid URL encoding at the HTTP Store part. While this is the correct part for the index, it might cause issues with the URI generated (say a file named `build+1.2.3.dll` for versioned files like our nupkgs do).

src/Microsoft.SymbolStore/SymbolStores/HttpSymbolStore.cs
mikem8361: EscapeDataString escapes too much especially the "/" the parts of the index. The difference between doing the escaping here and in BuildKey is that this code is escaping the whole index and the BuildKey code was escaping only the file name portion.

src/Microsoft.SymbolStore/KeyGenerators/KeyGenerator.cs
mikem8361: The uploader treats the index opaquely and it shouldn't be escaped. The symbol server was incorrectly indexing the file with the escaped index and the downloaders (windbg or dotnet-symbol) couldn't find it.  

I don't think it is a problem with the difference between EscapeDataString and EscapeUriString (:/?) when the index does get sent by the downloaders using symstore.

mikem8361:
I had the Edge team that request this fix test SymUploader end-to-end by uploading a file with blanks in the name and download with windbg.

And another point that makes this change not risky is that I've never seen a .NET Core module/symbol file that needs to be escaped.  I ran some local testing on various .NET Core symbol packages to confirm.

src/Microsoft.SymbolStore/KeyGenerators/KeyGenerator.cs
hoyosjs: > The uploader treats the index opaquely and it shouldn't be escaped. The symbol server was incorrectly indexing the file with the escaped index.

This is indeed what should happen for the key. We should never touch the file name for the key-building. What I am raising up here is that the behavior has changed, and that we could now build malformed URIs (especially on Unix). That being said, I'm OK with this as is.

mikem8361: I think I did find a problem that you are describing testing on this windows dump with the module "xd (dev).exe" in it.  The new code created the index: `xd (dev).exe/5d5c549511d9b000/xd (dev).exe` but the () didn't get escaped when the URI request was sent `http://msdl.microsoft.com/download/symbols/xd%20(dev).exe/5d5c549511d9b000/xd%20(dev).exe`

There is also a SymbolStoreKey.IsKeyValid function that the cache/directory symbol stores use that needs to be changed to allow more special characters that were escaped before.

mikem8361: This would be fixed by using EscapeDataString except for the part that is escapes "/".

mikem8361: Maybe escaping "/" is ok because it will be unescaped back to / on the symbol server before it looks the index up?

hoyosjs: I don't know how symbol servers do request routing. However, my guess is no. And as you said, I constructed this test: `foo&bar (1.3.preview+sha256)#hello.dll`. Generated an invalid URI before and after. Looks like a space before a parenthesis is doing weird things in URI encoding...

In you case, I see `xd (dev)` generating `"xd%20%28dev%29.exe/AAAAAAAAAAAAAA/xd%20%28dev%29.exe"`, but then the URI encoding does `{http://msdl.microsoft.com/download/symbols/xd %28dev%29.exe/AAAAAAAAAAAAAA/xd %28dev%29.exe}`

src/Microsoft.SymbolStore/SymbolStoreKey.cs
```diff
@@ -104,7 +107,7 @@ public static bool IsKeyValid(string index)
                     if (char.IsLetterOrDigit(c)) {
                         continue;
                     }
-                    if (c == '_' || c == '-' || c == '.' || c == '%') {
+                    if (!s_invalidChars.Contains(c)) {
```
hoyosjs: Not sure this will work as what we thought. The dictionary returns a bigger set on windows than on Linux. Linux files are allowed to have such characters, and we index our symbols on a Windows leg. Not sure if the symbol server can store this files, but the key creation should probably not fail in such case.  I consider what you are doing here an improvement nonetheless as at least this now allows to index all unicode named files.

Form the documentation:
> The array returned from this method is not guaranteed to contain the complete set of characters that are invalid in file and directory names. The full set of invalid characters can vary by file system. For example, on Windows-based desktop platforms, invalid path characters might include ASCII/Unicode characters 1 through 31, as well as quote ("), less than (<), greater than (>), pipe (|), backspace (\b), null (\0) and tab (\t).

mikem8361: The difference between Windows/Linux/MacOS might be what  we want because this function is only used by the Cache and DirectorySymbolStores which are writing files based on parts of the index to disk. If IsKeyValid fails, the cache or directory entry will be skipped.

PR #173, opened 2020-03-09
Minor eng cleanup

mikem8361:
Fix SourceLink validation failures.

Add back the PR merge build trigger

Add back setting DotNetPublishUsingPipelines

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -1,4 +1,5 @@
-trigger: none
+trigger:
+- master
 
 pr:
   autoCancel: true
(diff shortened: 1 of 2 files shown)


mikem8361:
An example of a successful official build: 

https://dev.azure.com/dnceng/internal/_build/results?buildId=552988&view=results

azure-pipelines.yml
```diff
@@ -10,18 +11,14 @@ pr:
...
     - THIRD-PARTY-NOTICES.TXT
-    - LICENSE.TXT
+    - License.txt
 
 variables:
   - name: _TeamName
     value: DotNetCore
-  - name: _DotNetArtifactsCategory
-    value: .NETCore
+  - name: _SignType
```
hoyosjs: This should not be set in public.

mikem8361: Why would setting it matter when _SignType is only used by the internal builds?

hoyosjs: Not a strong stance. I just don't like setting stuff that could get accidentally used in a refactor.

PR #174, opened 2020-03-10
Limit publishing to internal

hoyosjs:


mikem8361:
Can you also fix line 53 or whatever line causes the public CI to fail on PR merge.

```
-            ${{ if or(eq(variables['System.TeamProject'], 'public'), in(variables['Build.Reason'], 'PullRequest')) }}:
+            ${{ if eq(variables['System.TeamProject'], 'public') }}:
```

hoyosjs:
Actually, I misread. The CI  kickoff you sent was in public. This should run on the public pool and that's an `or` clause. That'd work as it is right now. What was failing was this job trying to publish on non-PR in public.

PR #188, opened 2020-04-23
Skip PDBs with empty DbiStreams

mikem8361:
Issue: https://github.com/dotnet/arcade/issues/4763

### DIFF
--- a/src/Microsoft.SymbolStore/KeyGenerators/PDBFileKeyGenerator.cs
+++ b/src/Microsoft.SymbolStore/KeyGenerators/PDBFileKeyGenerator.cs
@@ -33,7 +33,10 @@ namespace Microsoft.SymbolStore.KeyGenerators
             {
                 if ((flags & KeyTypeFlags.IdentityKey) != 0)
                 {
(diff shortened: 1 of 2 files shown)


src/Microsoft.SymbolStore/KeyGenerators/PDBFileKeyGenerator.cs
```diff
@@ -33,7 +33,10 @@ public override IEnumerable<SymbolStoreKey> GetKeys(KeyTypeFlags flags)
             {
                 if ((flags & KeyTypeFlags.IdentityKey) != 0)
                 {
-                    yield return GetKey(_path, _pdbFile.Signature, unchecked((int)_pdbFile.DbiAge));
+                    if (_pdbFile.DbiStream.IsValid())
+                    {
+                        yield return GetKey(_path, _pdbFile.Signature, unchecked((int)_pdbFile.DbiAge));
```
hoyosjs, line 38: Will not generating a key for a generator result in an exception somewhere, or will uptaking this in the uploader be safe?

mikem8361: No exception. It just doesn't add it to the list of keys to be returned to the symbol uploader.

PR #198, opened 2020-06-04
Remove named .gnu.debug.build-id section assumption

### LINKED ISSUE #197: Build-ID is forced to be in section named .gnu.debug.build-id
The current code to parse the build id from an ELF file assumes the build-id resides in a section name `.gnu.debug.build-id`

https://github.com/dotnet/symstore/blob/671770f5988345fae173cf62c662ce592fa0d9bd/src/Microsoft.FileFormats/ELF/ELFFile.cs#L177-L184

This doesn't necessarily have to be true. All that is required (both according to [Fedora's definition](https://fedoraproject.org/wiki/Releases/FeatureBuildId) of the build-id and the SSQP Key Convention is that

> The key is computed by reading the 20 byte sequence of the ELF Note section that is named “GNU” and that has note type PRPSINFO (3).

schultetwin1:
The code to read build-ids from ELF files assumes the build-id is in a
section named .gnu.debug.build-id. This doesn't necessarily have to be
true. In many firmware projects, the build-id needs to be placed
somewhere in particular and so a different section name is used.

Also, the SSQP protocol document doesn't defined a restriction on the
section name either. It just states:

> ELF Note section that is named “GNU” and that has note type
> PRPSINFO (3)

Which is still validated

This fixes #197


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/dotnet/symstore?pullRequest=198) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/dotnet/symstore?pullRequest=198) before we can accept your contribution.<br/><br/>:x: schultetwin1 [sign now](https://cla.dotnetfoundation.org/dotnet/symstore?pullRequest=198)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/dotnet/symstore?pullRequest=198) it.</sub>

PR #200, opened 2020-06-04
Inspect ELF to determine if symbol file

### LINKED ISSUE #199: ELF file extension does not always indicate debug info existance
Currently symstore uses the existence of a ".dbg" or ".debug" extension to indicate that a ELF file has debug symbols. This may not be the case. Especially for ELF files which were compiled with -g and never stripped.

Instead, we should check for the existence of debugging information in the ELF sections

schultetwin1:
When generating the Key for an ELF, the ELF-buildid-sym format

> applies to any ELF format files that have not been stripped of
> debugging information

This means that any ELF file, regardless of it's extension, should
generate a ELF-buildid-sym key if it has debugging symbols in it.

This changes the logic to detect if symbols exist or not by looking for
the .debug_info section.

This fixes #199


schultetwin1:
> as long as this always works for .NET Core binaries

I'm probably not the best person to speak to this, but as long as .NET isn't doing anything wonky with their debug information, this should work. Maybe there is someone more familiar with how the .NET team generates Linux binaries who could pitch in here?

In general, this is how ELF files store DWARF symbols. I did think of one (more recent) scenario called [split dwarf](https://gcc.gnu.org/wiki/DebugFission) in which your symbols file is a separate file (kind of like a PDB). In that case the section name is `.debug_info.dwo`. I think what I'll do is use the same logic the `strip` tool uses and just look for any section that starts with `.debug_` or `.zdebug_`

mikem8361:
.NET Core doesn't do any funky with our binaries and symbol files so all of this will be fine.  I was letting you know the history of this code which started out specifically only for .NET Core.

src/Microsoft.SymbolStore/KeyGenerators/ELFFileKeyGenerator.cs
```diff
@@ -62,7 +62,7 @@ public override IEnumerable<SymbolStoreKey> GetKeys(KeyTypeFlags flags)
                     bool symbolFile = false;
                     try
                     {
-                        symbolFile = Array.Exists(_elfFile.Sections, section => ((section.Name == ".debug_info") || (section.Name == ".zdebug_info")));
+                        symbolFile = Array.Exists(_elfFile.Sections, section => (section.Name.StartsWith(".debug_") || section.Name.StartsWith(".zdebug_")));
```
mikem8361, line 65: Not sure how important this is but there are a lot of other sections that begin with .debug_/.zdebug_ that I assume are only in a binary containing symbols but wouldn't doing `.StartsWith(".debug_info")/.StartsWith(".zdebug_info")` be more precise and still cover this case?

schultetwin1: Yep, I agree, we can err on the side of caution.

mikem8361:
@schultetwin1 are your symstore PRs ready to merge?

schultetwin1:
Yep! All good on my end. Thanks for the quick response and taking a look at them!

PR #201, opened 2020-06-04
Support 16 byte Build IDs

### LINKED ISSUE #196: ELF Key Conventions assume a 20-byte Build-ID
In the spec for [ELF Key Conventions](https://github.com/dotnet/symstore/blob/master/docs/specs/SSQP_Key_Conventions.md#elf-buildid) the build-id is specified as 20-bytes.

> This applies to any ELF format files that have not been stripped of debugging information, commonly ending in ‘.so.dbg’ or ‘.dbg’. The key is computed by reading the **20 byte sequence** of the ELF Note section that is named “GNU” and that has note type PRPSINFO (3). The file name is not used in the index because there are cases where all we have is the build id. The final key is formatted

However, the build-id is only 20 bytes if using `--build-id=sha1`. If using `--build-id=md5` or `--build-id=uuid` then it's only 16 bytes. More info [here](https://linux.die.net/man/1/ld).

The key convention should specify what to do in such a scenario

schultetwin1:
With Linux Build IDs, the developer can choose md5, uuid, sha1, or some
custom value. The SSQP protocol calls for a 20 byte value for a build-id
but md5 and uuid are only 16 bytes. This will pad 0x00s to the build-id
to reach a size of 20 bytes

This fixes #196


src/Microsoft.SymbolStore/KeyGenerators/ELFFileKeyGenerator.cs
```diff
@@ -57,6 +57,15 @@ public override IEnumerable<SymbolStoreKey> GetKeys(KeyTypeFlags flags)
             if (IsValid())
             {
                 byte[] buildId = _elfFile.BuildID;
+                if (buildId != null && buildId.Length < 20 && buildId.Length >= 16)
```
mikem8361, line 60: I think this should be in the below static GetKeys() functions and remove the length check here.  That allows the build id to be normalized in all cases. The static GetKeys() is used by some of tools (dotnet-dump) directly with the build id which also could be the 16 byte/md5 version.

schultetwin1: Ah, good call. I think we need to do this in both places then because of the calls to `BuildKey` later on in this function.

schultetwin1:
Fixing the conflicts

schultetwin1:
Done!

PR #216, opened 2020-08-31
Change PrpsInfo to GnuBuildId

mjsabby:
The ELF note headers give name "GNU" and type 3 (NT_GNU_BUILD_ID) for a build ID note, not PRPSINFO.


mjsabby:
cc @mikem8361 @janvorli @noahfalk

Issue #220, opened 2020-09-23
PE-timestamp-filesize format


DJm00n:
https://github.com/dotnet/symstore/blob/ccfa77ef0d395362d70a84404c24c8f30e81f039/docs/specs/SSQP_Key_Conventions.md#L15-L21

It's not clear if Timestamp should be extended to 8 chars with leading zeroes.

For example I have dll with [strange old timestamp](https://devblogs.microsoft.com/oldnewthing/20180103-00/?p=97705#:~:text=It's%20a%2032%2Dbit%20value,completely%20unrelated%20to%20any%20timestamp.):
![image](https://user-images.githubusercontent.com/1285934/94012623-b7a05c00-fdb1-11ea-93df-94f624a56b7a.png)

And for it path like`filename.dll/00E4915Eae000/filename.dll` is generated in MSVS request to symbol server.

But current `Microsoft.SymbolStore` implementation just generates PE file key without leading zeroes:
https://github.com/dotnet/symstore/blob/22c694a8c3217395c62ebaba523f2cc3dac775ba/src/Microsoft.SymbolStore/KeyGenerators/PEFileKeyGenerator.cs#L201

Seems like spec and code should be updated to match `DbgHelp.dll` implementation.

PS: https://randomascii.wordpress.com/2013/03/09/symbols-the-microsoft-way/ have note about that:
```
Note that the timeStamp is always printed as eight digits (with leading zeroes as needed) using upper-case for ‘A’ to ‘F’ (important if your symbol server is case sensitive), whereas the image size is printed using as few digits as needed, in lower-case.
```

DJm00n:
I think code should be changed to:
```
string id = string.Format("{0:X8}{1:x}", timestamp, sizeOfImage); 
```

tmat:
Note that when a module is build as reproducible the timestamp is not an actual time, but rather a content hash and hence could look odd when interpreted as time.

mikem8361:
It looks like we did get the DLL timestamp formatting wrong. It has been that way from the beginning of the symstore repo.   I want to verify this in the Windows debugger code that builds DLL indexes (I think it is symsrv) to confirm this.  We probably have gotten away with it because the timestamp for the .NET Core DLLs seems to never have any leading zeros and our servers are case-insensitive.  Hopefully because of this, making this change doesn't break our current upload, download, SOS and other tool usage. 

Adding @noahfalk and @tommcdon.

noahfalk:
[...]

PR #222, opened 2020-09-24
Fix the PE index formatting

mikem8361:
Symstore PE index formatting has been slightly incorrect since the beginning. We probably have gotten away with it because the
timestamp for the .NET Core DLLs seems to never have any leading zeros and our servers are case-insensitive. Hopefully because
of this, making this change doesn't break our current upload, download, SOS and other tool usage.

Issue: https://github.com/dotnet/symstore/issues/220

### DIFF
--- a/docs/specs/SSQP_Key_Conventions.md
+++ b/docs/specs/SSQP_Key_Conventions.md
@@ -20,6 +20,8 @@ This key references Windows Portable Executable format files which commonly have
 
 `<filename>/<Timestamp><SizeOfImage>/<filename>`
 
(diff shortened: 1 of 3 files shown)


noahfalk (approved):
LGTM

It is a little suspicious that the same formatting operation shows up three different times in the same assembly - it suggests the factoring might be inefficient. I didn't spend any time investigating the rationale or seeking a possible improvement because this stuff probably changes too rarely to be worth the effort.

mikem8361:
That was my thinking too.  I don't like the duplication but it is simpler to leave it alone.

PR #226, opened 2020-10-08
Fix PE-file key casing

ww898:
[SSQP_Key_Conventions.md](https://github.com/dotnet/symstore/blob/master/docs/specs/SSQP_Key_Conventions.md) require lower case everywhere.


ww898:
Sorry, haven't seen:h
> Note that the timeStamp is always printed as eight digits (with leading zeroes as needed) using upper-case for ‘A’ to ‘F’ (important if your symbol server is case sensitive), whereas the image size is printed using as few digits as needed, in lower-case.

PR #227, opened 2020-10-08
Fixed the example in the documentation

ww898:
Fixed the example in the documentation to accentuate the casing of lookup keys for PE-files.


ww898:
Is any body here to apply this change?

PR #238, opened 2020-11-07
Added Macho FAT header support

mikem8361:


src/Microsoft.SymbolStore/KeyGenerators/FileKeyGenerator.cs
```diff
@@ -53,6 +53,7 @@ private IEnumerable<KeyGenerator> GetGenerators()
                 yield return new MinidumpKeyGenerator(Tracer, _file);
                 yield return new ELFFileKeyGenerator(Tracer, _file);
                 yield return new PEFileKeyGenerator(Tracer, _file);
+                yield return new MachOFatHeaderKeyGenerator(Tracer, _file);
                 yield return new MachOFileKeyGenerator(Tracer, _file);
```
hoyosjs, line 57: Are MachO fat and regular file different such that the key generators won't overlap?

hoyosjs: On a second read, it looks like if it's a flat file, the FAT one would return null and the other one returns the specific stream. Is this correct?

src/Microsoft.FileFormats/ELF/ELFProgramHeader.cs
```diff
@@ -20,6 +15,17 @@ public enum ELFProgramHeaderType : uint
         Phdr = 6
     }
 
+    [Flags]
+    public enum ELFProgramHeaderFlags : uint
```
hoyosjs, line 19: Is this used anywhere at all?

src/Microsoft.FileFormats/ELF/ELFCoreFile.cs
```diff
@@ -30,6 +28,10 @@ public bool IsValid()
             return _elf.IsValid() && _elf.Header.Type == ELFHeaderType.Core;
         }
 
+        public bool Is64Bit => _elf.Is64Bit;
```
hoyosjs, line 31: Are these properties used?

hoyosjs (approved):

src/Microsoft.SymbolStore/KeyGenerators/FileKeyGenerator.cs
mikem8361: Yes, the two formats have different header signatures which causes one to fail (return null) and the other to succeed.

src/Microsoft.FileFormats/ELF/ELFCoreFile.cs
mikem8361: They will be.  I needed for this little dump tool that displays info about ELF dumps.

src/Microsoft.FileFormats/ELF/ELFProgramHeader.cs
mikem8361: This will be used in some upcoming diagnostics extensions work. I wanted to get them in now.

Issue #256, opened 2021-02-08
Dotnet-Symbols Failures Should Display Which Files Failed


brianrob:
When attempting to download symbol files using `dotnet-symbols`, if the files don't exist, the user gets something that looks like the following:

```
ERROR: HttpSymbolStore: 404 Not Found 'http://msdl.microsoft.com/download/symbols/_.debug%2Felf-buildid-sym-ae6cda7c380adee16ca6f69db1d112a35030d2dc%2F_.debug'
ERROR: HttpSymbolStore: 404 Not Found 'http://msdl.microsoft.com/download/symbols/_.debug%2Felf-buildid-sym-e182f3ef12199c868f4829168e8cc6049a72bc18%2F_.debug'
ERROR: HttpSymbolStore: 404 Not Found 'http://msdl.microsoft.com/download/symbols/_.debug%2Felf-buildid-sym-d1eaaae4c0e7206d3d9cb5c291ca5b47a6366cab%2F_.debug'
ERROR: HttpSymbolStore: 404 Not Found 'http://msdl.microsoft.com/download/symbols/_.debug%2Felf-buildid-sym-00add623781f2dfc54e25116f2221cfd5dd65350%2F_.debug'
ERROR: HttpSymbolStore: 404 Not Found 'http://msdl.microsoft.com/download/symbols/_.debug%2Felf-buildid-sym-d65f89c92f7f3be3b4e1e4334549f01b366ea17a%2F_.debug'
ERROR: HttpSymbolStore: 404 Not Found 'http://msdl.microsoft.com/download/symbols/_.debug%2Felf-buildid-sym-6be8be1359f4b49868d19681be8eaf6726293d84%2F_.debug'
ERROR: HttpSymbolStore: 404 Not Found 'http://msdl.microsoft.com/download/symbols/_.debug%2Felf-buildid-sym-9f2924424d15f438a6e527f48003e001893ad988%2F_.debug'
ERROR: HttpSymbolStore: 404 Not Found 'http://msdl.microsoft.com/download/symbols/_.debug%2Felf-buildid-sym-dcd7052cd56d6047c98777d9797da87d51757ba3%2F_.debug'
ERROR: HttpSymbolStore: 404 Not Found 'http://msdl.microsoft.com/download/symbols/_.debug%2Felf-buildid-sym-76555729e0b2399645e221227b872d20d137e462%2F_.debug'
[...]

Issue #246, opened 2020-12-16
Download is slow; diagnostic tools are unusable


am11:
When I download a .NET SDK tarball from http://aka.ms/dotnet/net5/5.0.1xx/daily/Sdk/dotnet-sdk-linux-x64.tar.gz (133 MB), it is served via Verizon CDN, and hence results in a decent transmission rate: 
```sh
$ wget http://aka.ms/dotnet/net5/5.0.1xx/daily/Sdk/dotnet-sdk-linux-x64.tar.gz
...
HTTP request sent, awaiting response... 200 OK
Length: 139267404 (133M) [application/octet-stream]
Saving to: ‘dotnet-sdk-linux-x64.tar.gz’

dotnet-sdk-linux-x64.tar.gz                       
100%[======================================================================================
=============================>] 132.82M  3.24MB/s    in 49s     

2020-12-16 06:54:48 (2.69 MB/s) - ‘dotnet-sdk-linux-x64.tar.gz’ saved [139267404/139267404]
```

However, downloading a smaller blob from symstore via `dotnet symbol -d singlefilehost` (113 MB) times out after 4 minutes.

```sh
$ dotnet symbol -d singlefilehost
Downloading from http://msdl.microsoft.com/download/symbols/
ERROR: The operation was canceled. -> Unable to read data from the transport connection: Operation canceled.
```

`httpry(1)` shows the URL to file:
```sh
$ sudo httpry -i enp0s8
...
2020-12-16 07:01:31	192.168.1.224	204.79.197.219	>	GET	msdl.microsoft.com	/download/symbols/_.debug%2Felf-buildid-sym-6ca69f7bc7061d8eab20013d6820ef5a5c2f7ff4%2F_.debug	HTTP/1.1	-
2020-12-16 07:01:32	204.79.197.219	192.168.1.224	<	-	-	-	HTTP/1.1	302	Found
```
tried that URL with `wget(1)`:
```sh
$ wget http://msdl.microsoft.com/download/symbols/_.debug%2Felf-buildid-sym-6ca69f7bc7061d8eab20013d6820ef5a5c2f7ff4%2F_.debug
...
HTTP request sent, awaiting response... 200 OK
Length: 118073560 (113M) [application/octet-stream]
Saving to: ‘_.debug%2Felf-buildid-sym-6ca69f7bc7061d8eab20013d6820ef5a5c2f7ff4%2F_.debug’

[...]

PR #257, opened 2021-02-11
Add --timeout dotnet-symbol command line option and better "Not Found" error message

mikem8361:
Make reading section name more bullet proof

Issue: https://github.com/dotnet/symstore/issues/256 and https://github.com/dotnet/symstore/issues/246.

### DIFF
--- a/src/dotnet-symbol/Program.cs
+++ b/src/dotnet-symbol/Program.cs
@@ -108,7 +108,7 @@ namespace dotnet.symbol
                     case "--timeout":
                         if (++i < args.Length)
                         {
-                            int timeoutInMinutes = int.Parse(args[i]);
+                            double timeoutInMinutes = double.Parse(args[i]);
                             program.Timeout = TimeSpan.FromMinutes(timeoutInMinutes);
                         }
                         else 
--- a/src/dotnet-symbol/Program.cs
+++ b/src/dotnet-symbol/Program.cs
@@ -27,6 +27,7 @@ namespace dotnet.symbol
         private readonly List<string> CacheDirectories = new List<string>();
         private readonly List<ServerInfo> SymbolServers = new List<ServerInfo>();
         private string OutputDirectory;
+        private TimeSpan? Timeout;
         private bool Subdirectories;
         private bool Symbols;
         private bool Debugging;
@@ -96,6 +97,16 @@ namespace dotnet.symbol
                         if (++i < args.Length)
                             program.OutputDirectory = args[i];
                         else
+                            goto usage;
+                        break;
+
+                    case "--timeout":
+                        if (++i < args.Length)
+                        {
+                            int timeoutInMinutes = int.Parse(args[i]);
+                            program.Timeout = TimeSpan.FromMinutes(timeoutInMinutes);
+                        }
+                        else 
                             goto usage;
                         break;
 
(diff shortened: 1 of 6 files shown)


am11:
Thanks a lot Mike! :)
The second part of the issue is that the symbol blobs appear to be stored in cold/cheap tier storage, hence making them inaccessible (or super slow transfer) at times in EU regions; would help if they are moved under the same plan as .NET installer blobs. Maybe it is just missing CDN configuration. This slow transmission rate can also be observed in a VM deployed in North Ireland or Netherlands Azure datacenters.

hoyosjs:
> The second part of the issue is that the symbol blobs appear to be stored in cold/cheap tier storage, hence making them inaccessible (or super slow transfer) at times in EU regions; would help if they are moved under the same plan as .NET installer blobs. Maybe it is just missing CDN configuration. This slow transmission rate can also be observed in a VM deployed in North Ireland or Netherlands Azure datacenters.

Sadly this is not something we control. The symbol servers for the company run on a type of 2 tier system that is not particularly the same as a blob storage account as you see for the SDK. That one has a georeplication and some other caching mechanisms. I might reach out to them again, but last time I did there was still not much of a plan around that area.

src/dotnet-symbol/Program.cs
```diff
@@ -99,6 +101,20 @@ public static void Main(string[] args)
                             goto usage;
                         break;
 
+                    case "--overwrite":
+                        program.Overwrite = true;
+                        break;
+
+                    case "--timeout":
```
hoyosjs, line 108: Maybe add the units to the flag itself. Also, should we use a double instead of an int? `TimeSpan.FromMinutes` takes one.

src/dotnet-symbol/Program.cs
```diff
@@ -355,9 +375,9 @@ private async Task WriteFileToDirectory(Stream stream, string fileName, string d
         {
             stream.Position = 0;
             string destination = Path.Combine(destinationDirectory, Path.GetFileName(fileName.Replace('\\', '/')));
-            if (File.Exists(destination))
+            if (!Overwrite && File.Exists(destination))
```
hoyosjs, line 378: Two things I never considered about this:

- This won't work well for things like ALCs, where multiple versions of a dll with the same name will end up with symbol collisions.
- With this flag, we should pull this logic up to `DownloadFiles` and only download the file if (Overwrite || !File.Exists(key.FullPathName))

hoyosjs (approved):

src/dotnet-symbol/Program.cs
mikem8361: The help makes it clear that it is in minutes.  I hope that is enough. I would like keep the option name simple. I'll change it to a double.

src/dotnet-symbol/Program.cs
mikem8361: Since the default is still NOT to overwrite, your first point isn't any worse. If we have name collisions then the first one wins which is bad (already bad).

Your second point is a good idea. I'll try it out.

mikem8361: Never mind about the second part.  It would involve too many changes and refactor to get the "destination" file path in download.  I hope it is ok to leave it be.

PR #258, opened 2021-02-13
Use EscapeUriString instead of EscapeDataString

am11:
It seems like due to the recent change at the storage side, the escaped `index` component of URL is not working anymore.

#### Before

```sh
$ dotnet symbol --symbols mysinglefilehostapp1 --timeout 20
Downloading from http://msdl.microsoft.com/download/symbols/
ERROR: Not Found: singlefilehost.dbg - 'http://msdl.microsoft.com/download/symbols/_.debug%2Felf-buildid-sym-27997343859af70e0bfc5e50a4d1059b3513a1c4%2F_.debug'
```

http://msdl.microsoft.com/download/symbols/_.debug%2Felf-buildid-sym-27997343859af70e0bfc5e50a4d1059b3513a1c4%2F_.debug is returning 404.

Replacing `%2F` with `/` fixes the issue: http://msdl.microsoft.com/download/symbols/_.debug/elf-buildid-sym-27997343859af70e0bfc5e50a4d1059b3513a1c4/_.debug is valid.

#### After

```sh
$ dotnet symbol --symbols mysinglefilehostapp1 --timeout 20
Downloading from http://msdl.microsoft.com/download/symbols/
Writing: ./singlefilehost.dbg
```

To reproduce it on machine with pre-downloaded symbol, rename `~/.dotnet/symbolcache` directory temporarily to force the re-download (or try it in a new VM).

Fixes https://github.com/dotnet/diagnostics/issues/1978.

### DIFF
--- a/src/Microsoft.SymbolStore/SymbolStores/HttpSymbolStore.cs
+++ b/src/Microsoft.SymbolStore/SymbolStores/HttpSymbolStore.cs
@@ -119,7 +119,7 @@ namespace Microsoft.SymbolStore.SymbolStores
 
         protected Uri GetRequestUri(string index)
         {
-            if (!Uri.TryCreate(Uri, Uri.EscapeDataString(index), out Uri requestUri))
+            if (!Uri.TryCreate(Uri, Uri.EscapeUriString(index), out Uri requestUri))
             {
                 throw new ArgumentException(nameof(index));
             }


src/Microsoft.SymbolStore/SymbolStores/HttpSymbolStore.cs
```diff
@@ -119,7 +119,7 @@ protected override async Task<SymbolStoreFile> GetFileInner(SymbolStoreKey key,
 
         protected Uri GetRequestUri(string index)
         {
-            if (!Uri.TryCreate(Uri, Uri.EscapeDataString(index), out Uri requestUri))
```
mikem8361, line 122: I moved the index escaping from the key generations in KeyGenerator.cs to here about a year ago in PR #170. I don't think the fix is as simple as removing the escaping here.  Need to do more investigation on why this is happening now.

am11: True, it was working with escaping until last week of January at least (last time I tried on a fresh instance). I think this is due to some change on the storage side. It also redirects from http to https now (which happens seamlessly).

mikem8361: In the end, we still need to escape spaces and other invalid characters.  We (all the teams at MS that use the uploader and downloader based on symstore) do have files that have spaces in them that we need to download.  

I'm going to suggest that we use Uri.EscapeUriString here.  There is some controversy (see the last paragraph [here](https://stackoverflow.com/questions/4396598/whats-the-difference-between-escapeuristring-and-escapedatastring)) about this API vs Uri.EscapeDataString. EscapeUriString escapes only illegal/invalid characters like space where EscapeDataString also escapes "reserved" chars like "/".  EscapeUriString has the downside that it can double escape already escaped characters but that won't happen here because "index" is never already "escaped". 

/cc: @hoyosjs

hoyosjs:
Hmm. Let me think about this, because this codepath gets used by sos, dotnet-dump, the symbol uploader, and dotnet-symbol. I'd like to verify that it's compatible with all the versions that we have used to upload product bits. If anything, I'm suspecting the metadata store for symbols wasn't all that smart and we were just matching upload and download. I'll need to take a quick look when I can.

src/Microsoft.SymbolStore/SymbolStores/HttpSymbolStore.cs
am11: Makes sense. Using `EscapeUriString` for `index` also worked:

Original: `_.debug/elf-buildid-sym-27997343859af70e0bfc5e50a4d1059b3513a1c4/_.debug`
EscapeUriString : `_.debug/elf-buildid-sym-27997343859af70e0bfc5e50a4d1059b3513a1c4/_.debug`
EscapeDataString: `_.debug%2Felf-buildid-sym-27997343859af70e0bfc5e50a4d1059b3513a1c4%2F_.debug`

mikem8361:
@hoyosjs the more I think about this the more I think using Uri.EscapeUriString is the right thing to do.  It won't affect the symbol uploader.  And it allow dotnet-symbol, SOS/dotnet-dump to work with those 5.0.x, etc. builds.

mikem8361:
@am11, I put up PR #260 that fixes the problems with EscapeUriString and doesn't escape forward slashes.

PR #260, opened 2021-02-24
Don't escape forward slashes in the index portion of the HTTP URL

mikem8361:


src/Microsoft.SymbolStore/SymbolStores/HttpSymbolStore.cs
```diff
@@ -119,7 +119,9 @@ protected override async Task<SymbolStoreFile> GetFileInner(SymbolStoreKey key,
 
         protected Uri GetRequestUri(string index)
         {
-            if (!Uri.TryCreate(Uri, Uri.EscapeUriString(index), out Uri requestUri))
+            // Escape everything except the forward slashes (/) in the index
+            index = string.Concat(index.Split('/').Select((part, index) => (index == 0 ? "" : "/") + Uri.EscapeDataString(part)));
```
hoyosjs, line 123: Kinda of sad to do this leaky abstraction here, but not sure if we can do much without investing a lot of effort...

This one will allocate a little less:

```suggestion
            index = string.Join("/", index.Split('/').Select(part => Uri.EscapeDataString(part)));
```

hoyosjs (approved):

src/Microsoft.SymbolStore/SymbolStores/HttpSymbolStore.cs
mikem8361: Yes, that is a lot better way to do this.

PR #302, opened 2021-09-29
Add reading PE export symbols to PEFile

mikem8361:


src/Microsoft.FileFormats/PE/PEFile.cs
```diff
@@ -105,6 +106,55 @@ public bool IsValid()
...
+            }
+            offset = 0;
+            return false;
+        }
+
+        private ImageExportDirectory ReadExportDirectory()
+        {
+            if (IsValid())
+            {
+                ImageDataDirectory exportTableDirectory = ImageDataDirectory[(int)ImageDirectoryEntry.Export];
```
hoyosjs, line 149: Is this resilient to only this section being available? I think PEFile expects to see all of the ImageDataDirectory, which is not even true in the PE spec for a regular image. (It assumes that always 15 sections will be available, completely ignoring `SizeOfHeaders`)

hoyosjs: Can we assert that the Characteristics == 0?

src/Microsoft.FileFormats/PE/PEFile.cs
```diff
@@ -105,6 +106,55 @@ public bool IsValid()
...
+                    for (int nameIndex = 0; nameIndex < exportDirectory.NumberOfNames; nameIndex++)
+                    {
+                        uint namePointerRVA = RelativeVirtualAddressReader.Read<uint>((ulong)(exportDirectory.AddressOfNames + (sizeof(uint) * nameIndex)));
+                        if (namePointerRVA != 0)
+                        {
+                            string name = RelativeVirtualAddressReader.Read<string>(namePointerRVA);
+                            if (name == symbolName)
+                            {
+                                ushort ordinalForNamedExport = RelativeVirtualAddressReader.Read<ushort>((ulong)(exportDirectory.AddressOfNameOrdinals + (sizeof(ushort) * nameIndex)));
+                                offset = RelativeVirtualAddressReader.Read<uint>((ulong)(exportDirectory.AddressOfFunctions + (sizeof(uint) * ordinalForNamedExport)));
```
hoyosjs, line 131: This does not consider the case of a forward RVA export, thought that is an edge case that might not apply to our case. If `offset` is within the export section, then according to https://docs.microsoft.com/en-us/windows/win32/debug/pe-format#export-address-table it's a pointer to a string that forwards the export to another dll potentially.

hoyosjs: Can we check `ordinalForNamedExport` < `NumberOfFunctions`?

hoyosjs (approved):

src/Microsoft.FileFormats/PE/PEFile.cs
mikem8361: The symstore PE reader has always assumed that there are 15 sections and that they all are in the dump. It reads all 15 in ReadImageDataDirectory().  It sounds like you think it is important enough to only read what fits in SizeOfHeaders?  Does SizeOfHeaders include the size of the file header (ImageFileHeader) and optional header (ImageOptionalHeader)?

Should it throw an exception or do Debug.Assert if Characteristics != 0? A debug symstore isn't ever really used.

src/Microsoft.FileFormats/PE/PEFile.cs
mikem8361: The runtime PEDecoder doesn't check for either a forward RVA export or if `ordinalForNamedExport < NumberOfFunctions`.

src/Microsoft.FileFormats/PE/PEFile.cs
hoyosjs: Not sure if throw bad image, or say I can't read (return null);

PR #321, opened 2022-01-18
Fix MacOS builds

mikem8361:


hoyosjs (approved):
Should we use macos-latest to avoid this breaking often? This is all managed so I don't expect issues.

Issue #327, opened 2022-02-10
Fix dotnet-symbol for newer MacOS dumps


mikem8361:
Fix the "dyld_all_images_infos" symbol lookup to try all the symbols instead of just the export symbols.

mikem8361:
Fixed in dotnet-symbol version 1.0.317802

PR #334, opened 2022-03-23
Fix issue https://github.com/dotnet/symstore/issues/327

mikem8361:
Reworked the symbol lookup code. Needed to fallback to search all symbols if it can find the it in the exported symbols. It always read every symbol name in the old path.


src/Microsoft.FileFormats/MachO/MachOFile.cs
```diff
@@ -448,28 +448,93 @@ public class MachSymtab
...
+            NList[] symTable = _symbolTable.Value;
+            if (symTable is not null)
+            {
+                for (uint i = 0; i < nsyms && start + i < symTable.Length; i++)
+                {
+                    string name = _stringReader.Value.Read<string>(symTable[start + i].StringIndex); 
+                    if (name.Length > 0)
+                    {
+                        // Skip the leading underscores to match Linux externs
+                        if (name[0] == '_')
```
hoyosjs, line 502: Future todo: this is going to allocate for all linux externs for every module. If we see this being an issue, use System.Memory + Span.Slice + SequenceEquals.

hoyosjs (approved):

PR #336, opened 2022-03-30
Add retry logic to HttpSymbolStore to help make the diagnostics repo tests more reliable

mikem8361:
Don't mark permanent failures for some socket errors.

### DIFF
--- a/src/Microsoft.SymbolStore/SymbolStores/HttpSymbolStore.cs
+++ b/src/Microsoft.SymbolStore/SymbolStores/HttpSymbolStore.cs
@@ -8,6 +8,7 @@ using System.Linq;
 using System.Net;
 using System.Net.Http;
 using System.Net.Http.Headers;
+using System.Net.Sockets;
 using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
@@ -46,6 +47,11 @@ namespace Microsoft.SymbolStore.SymbolStores
                 }
             }
         }
+
+        /// <summary>
+        /// The number of retries to do on a retryable status or socket error
+        /// </summary>
+        public int RetryCount { get; set; }
 
         /// <summary>
         /// Create an instance of a http symbol store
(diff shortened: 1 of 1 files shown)


src/Microsoft.SymbolStore/SymbolStores/HttpSymbolStore.cs
```diff
@@ -216,10 +250,19 @@ public override void Dispose()
...
-        protected bool IsRetryableStatus(HttpStatusCode status)
+        protected bool IsRetryableStatus(HttpStatusCode status) => s_retryableStatusCodes.Contains(status);
+
+        private HashSet<SocketError> s_retryableSocketErrors = new HashSet<SocketError>
         {
-            return s_retryableStatusCodes.Contains(status);
-        }
+            SocketError.ConnectionReset,
+            SocketError.ConnectionAborted,
+            SocketError.ConnectionRefused,
```
gregg-miskelly, line 259: Do we `ConnectionRefused` in the wild? I wonder if we should eventually call `MarkClientFailure();` if we keep hitting that since this the error code that we would get back if the symbol server exists in DNS but doesn't have an HTTP/HTTPS server?

mikem8361: I'm still figuring out the exactly list of retryable socket errors.   And yes I'll probably remove ConnectionRefused.

hoyosjs (approved):

src/Microsoft.SymbolStore/SymbolStores/HttpSymbolStore.cs
mikem8361: Are suggesting we do something IsConnectionResetError and IsConnectionAbortError together? Other than ConnectionRefused and the platform specific checks in those functions, what I have is pretty close.

Would removing ConnectionRefused be enough do you think?

PR #391, opened 2022-12-06
Fix ELF build id key generator padding

mikem8361:
8 byte/64 bit build ids need to be padding to 20 bytes/160 bits according to spec.

### DIFF
--- a/src/Microsoft.SymbolStore/KeyGenerators/ELFFileKeyGenerator.cs
+++ b/src/Microsoft.SymbolStore/KeyGenerators/ELFFileKeyGenerator.cs
@@ -58,8 +58,7 @@ namespace Microsoft.SymbolStore.KeyGenerators
             if (IsValid())
             {
                 byte[] buildId = _elfFile.BuildID;
-                NormalizeBuildId(ref buildId);
-                if (buildId != null)
+                if (NormalizeBuildId(ref buildId))
                 {
                     bool symbolFile = false;
                     try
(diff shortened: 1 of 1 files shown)


mikem8361:
@hoyosjs ping

src/Microsoft.SymbolStore/KeyGenerators/ELFFileKeyGenerator.cs
```diff
@@ -196,7 +196,7 @@ private string GetSymbolFileName()
         /// <returns>symbol store keys</returns>
         private static void NormalizeBuildId(ref byte[] buildId)
         {
-            if (buildId != null && buildId.Length < 20 && buildId.Length >= 16)
```
hoyosjs, line 199: This should perhaps throw an exception if buildID is greater than 20 bytes (and not sure why we don't do that in null cases), and remove the 8 byte check to be compliant with the spec.

mikem8361:
Push the some review feedback just in case we want to take this change.

src/Microsoft.SymbolStore/KeyGenerators/ELFFileKeyGenerator.cs
```diff
@@ -113,57 +112,59 @@ public override IEnumerable<SymbolStoreKey> GetKeys(KeyTypeFlags flags)
...
+                            {
+                                yield return BuildKey(specialFileName, CoreClrPrefix, buildId);
+                            }
                         }
                     }
                 }
             }
+            else
+            {
+                Debug.Fail($"Invalid ELF BuildId '{(buildId == null ? "<null>" : ToHexString(buildId))}' for {path}");
```
hoyosjs, line 166: Shouldn't we do an error log - it feels fair for this to not enumerate anything instead of failing on a debug build. Some files will likely have invalid index and we should just not generate an index for it under the current design.

mikem8361: We can't log anything to the tracer because this is a static entry point used by the diagnostics repo code and not the symstore key generators.  It previous asserted if the build id parameter was null or if it wasn't 20 bytes long so this is in keeping with the previous behavior and doesn't return any keys.

mikem8361: I was wrong about my that this static GetKeys function was only called by diagnostics repo.  It look like it is also called by the instance GetKeys method but after it's NormalizeBuildId call so the assert should never hit in that path.  The static method should not be called by the key generators directly or the uploader.

hoyosjs: Unless we explicitly make this private, you can't ensure that. What tells you symbol.exe wouldn't take a dependency on this? Why is this a necessary stati public method and not just always an instance method?

mikem8361: It is a public static entry point now so it can't be made private unless we want to break things. The diag repo code (and probably others) needs a key generator entry point that takes a build id, etc. to build the key from. The KeyGenerator instance classes are went to read the build id from the ELF or MachO binary. 

In the end, if the build id is invalid, the enumerable returned will have no items so that is a detectable failure.

hoyosjs (approved):

PR #405, opened 2023-01-30
Add support for non-Basic Auth PAT tokens

WardenGnaw:
This PR adds in a new class SymbolAuthHeader which contains a scheme and parameter that will be passed into AuthenticationHeaderValue's constructor.

This will allow non-"Basic" auth headers such as "Bearer" and etc.

@gregg-miskelly @chuckries

### DIFF
--- a/src/Microsoft.SymbolStore/SymbolStores/HttpSymbolStore.cs
+++ b/src/Microsoft.SymbolStore/SymbolStores/HttpSymbolStore.cs
@@ -15,6 +15,12 @@ using System.Threading.Tasks;
 
 namespace Microsoft.SymbolStore.SymbolStores
 {
+    public class SymbolAuthHeader
+    {
+        public string Scheme { get; set; }
+        public string Parameter { get; set; }
+    }
+
     /// <summary>
     /// Basic http symbol store. The request can be authentication with a PAT for VSTS symbol stores.
     /// </summary>
@@ -34,13 +40,13 @@ namespace Microsoft.SymbolStore.SymbolStores
         /// </summary>
         public TimeSpan Timeout
         {
-            get 
-            { 
+            get
+            {
                 return _client.Timeout;
             }
-            set 
-            { 
-                _client.Timeout = value; 
+            set
+            {
+                _client.Timeout = value;
                 if (_authenticatedClient != null)
                 {
                     _authenticatedClient.Timeout = value;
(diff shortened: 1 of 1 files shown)


chuckries (approved):

chuckries:
cc @mikem8361

src/Microsoft.SymbolStore/SymbolStores/HttpSymbolStore.cs
```diff
@@ -15,6 +15,12 @@
 
 namespace Microsoft.SymbolStore.SymbolStores
 {
+    public class SymbolAuthHeader
+    {
+        public string Scheme { get; set; }
+        public string Parameter { get; set; }
```
gregg-miskelly, line 21: Should we make this an immutable class that throws in the constructor if either of these is null?

gregg-miskelly: Actually, I wonder if you want to get rid of this type and instead just pass these as arguments. The reason I suggest this is that you now have two constructors with the same number of arguments. So if someone does `new HttpSymbolStore(tracer, backingStore, symbolServerUri, null)` it is ambiguous which constructor to call. Where as if you get rid of this type and just add two arguments to the constructor it is no longer ambiguous.

mikem8361: I like the idea of removing the SymbolAuthHeader class and passing the members as parameters to the new constructor.

mikem8361 (approved):

src/Microsoft.SymbolStore/SymbolStores/HttpSymbolStore.cs
```diff
@@ -69,26 +70,70 @@ public HttpSymbolStore(ITracer tracer, SymbolStore backingStore, Uri symbolServe
...
+        /// <param name="tracer">logger</param>
+        /// <param name="backingStore">next symbol store or null</param>
+        /// <param name="symbolServerUri">symbol server url</param>
+        /// <param name="authenticationHeader">The header information to use for the AuthenticationHeaderValue</param>
+        public HttpSymbolStore(ITracer tracer, SymbolStore backingStore, Uri symbolServerUri, string scheme, string parameter)
+            : this(tracer, backingStore, symbolServerUri, true)
+        {
+            if (string.IsNullOrEmpty(scheme))
+            {
+                throw new ArgumentException(nameof(scheme));
```
gregg-miskelly, line 123: >ArgumentException [](http://example.com/codeflow?start=26&length=17)

The single arg overload of `ArgumentException` accepts a message argument rather than an argument name. You want either `ArgumentOutOfRangeException` or `ArgumentNullException`

gregg-miskelly (approved):

mikem8361:
Ready to merge?

WardenGnaw:
> Ready to merge?

Yep! Thanks for the help @mikem8361

PR #407, opened 2023-02-10
Update README.md

DJm00n:
Add direct link to package

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,11 @@
 
 Implements API for retrieval of symbols and other debug artifacts from symbol store.
 
-Builds are available on the dotnet-tools feed: https://dev.azure.com/dnceng/public/_artifacts/feed/dotnet-tools/NuGet/Microsoft.SymbolStore
+Builds are available on the dotnet-tools feed:
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -2,7 +2,7 @@
 
 Implements API for retrieval of symbols and other debug artifacts from symbol store.
 
-Builds are available on the dotnet-tools feed: https://dev.azure.com/dnceng/public/_packaging?_a=feed&feed=dotnet-tools
+Builds are available on the dotnet-tools feed: https://dev.azure.com/dnceng/public/_artifacts/feed/dotnet-tools/NuGet/Microsoft.SymbolStore
```
mikem8361, line 5: This repo products more than just the Microsoft.SymbolStore package. Microsoft.FileFormats, dotnet-symbol, etc are also produced and published to that feed so it probably should be just: https://dev.azure.com/dnceng/public/_artifacts/feed/dotnet-tools

DJm00n: maybe you can list the packages in readme then? It just not obvious which package should I reference to use this lib.

mikem8361 (approved):

PR #409, opened 2023-02-17
Fix arithmetic overflow on dump Maoni found

mikem8361:


src/Microsoft.FileFormats/AddressSpaces.cs
```diff
@@ -66,7 +66,7 @@ public uint Read(ulong position, byte[] buffer, uint bufferOffset, uint count)
         /// <summary>
         /// The upper bound (non-inclusive) of readable addresses
         /// </summary>
-        public ulong Length { get { return (ulong)((long)_baseStart + (long)_length + _baseToRelativeShift); } }
+        public ulong Length { get { return unchecked(_baseStart + _length + (ulong)_baseToRelativeShift); } }
```
hoyosjs, line 69: I am not sure this is right. _baseToRelativeShift is potentially negative. Coercing it here to ulong could cause issues, couldn't it? I don't know why the construction validation is off. This is weird.`_baseStart + _baseToRelativeShift` might need to be done first and then add the length. Something like `((ulong)_baseToRelativeShift + _baseStart) + _length`. If that overflows, we likely have issues.

mikem8361: _baseToRelativeShift is negative but converted to ulong and adding it should be equivalent to the original code.  Without the unchecked this convert causes an overflow.  The original problem the overflow was caused by _length being 0xfffffffffff61000 which comes from the highest program header load address. 

It would also fix the original problem by just adding the unchecked around the original code.  I thought where I ended up was simpler with less casts but still equivalent.

mikem8361: Does this make sense to you? Isn't adding a ulong that is a negative long (_baseToRelativeShift is always negative) equivalent? It shouldn't (and doesn't) matter the order either.

hoyosjs (approved):

